PR #5, opened 2025-11-10
Windows compatibility

cclin130:
- `test_gcp.py` closes the access to tempfile before downloading, because Windows does not allow two open file handles
- use `posixpath` when declaring GCS URLs to ensure forward slashes. However, we still use `os.path.join` when declaring local directories because `os.path.join` uses the right forward vs backward slash depending on the OS

# Testing
` uv run pytest`
`uv run pytest -m ceda_integration`

Will test live with @bl-young before mergin

### DIFF
--- a/ceda_usa/utils/__tests__/test_gcp.py
+++ b/ceda_usa/utils/__tests__/test_gcp.py
@@ -1,17 +1,14 @@
 import os
-import tempfile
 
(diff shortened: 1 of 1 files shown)


bl-young:
The `test_gcp` is still giving my struggles with the `FileExistsError`. We can look closer this afternoon

PR #16, opened 2025-11-12
move ceda and flowsa under bedrock module

cclin130:
`uv` projects should be structured with the actual module dictated by the `pyproject.toml` in a directory below the main dir. This PR adds a top-level folder named `bedrock` and puts `ceda_usa` and `flowsa` under it.

- update import paths in all ceda_usa and flowsa files
- some flowsa files used `importlib.import_module()` and i updated the module path to include `bedrock`

I expect the `format` check in CI to fail and will let ben/catherine to address in a follow up PR

# Testing
I ran `uv run pytest` and the non-integration CEDA tests passed

i also ran successfully
```
import bedrock.flowsa


from bedrock.flowsa.generateflowbyactivity import main
main(source='EPA_GHGI_T_2_1', year=2022)
```

i will run CEDA integration tests on this branch
https://github.com/cornerstone-data/bedrock/actions/runs/19309486718

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
         uses: ./.github/actions/setup-python
 
       - name: Typecheck w/ mypy
-        run: uv run mypy ceda_usa
+        run: uv run mypy bedrock/ceda_usa
 
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


cclin130:
@bl-young im aligned, sorry for the huge PR

Issue #25, opened 2025-11-20
Write function to identify the most recent file(s) by name


bl-young:
This is analogous to [esupy.processed_data_mgmt.get_most_recent_from_index()](https://github.com/USEPA/esupy/blob/main/esupy/processed_data_mgmt.py#L158)

Issue #24, opened 2025-11-20
Write function to parse and sort all files on GCS at target path


bl-young:
Analogous to `esupy.processed_data_mgmt.get_data_commons_index()` and `.parse_data_commons_index()` [here](https://github.com/USEPA/esupy/blob/main/esupy/processed_data_mgmt.py).

PR #29, opened 2025-11-20
Update retrieve functions from GCP

### LINKED ISSUE #22: Write function to retrieve target file from GCS
- Pass name and path (directory to file on GCS)
- Download to local
- Include metadata file if any
- If multiple versions of the same file (e.g. different hash or version) obtain the most recent one

This will build on [download_gcs_file()](https://github.com/cornerstone-data/bedrock/blob/dc8d374354bb6b0e5e3651dcdb586165cb968306/bedrock/ceda_usa/utils/gcp.py#L55) and/or [load_from_gcs()](https://github.com/cornerstone-data/bedrock/blob/dc8d374354bb6b0e5e3651dcdb586165cb968306/bedrock/ceda_usa/utils/gcp.py#L32)

bl-young:
This PR updates the GCS retrieval process so that filenames that includes hash and version number can still be extracted when just the name is indicated as the target.

- The related functions are moved from `bedrock/ceda_usa/utils/gcp.py` to `bedrock/utils.gcp.py`
- Load `load_from_gcs()` is modified such that the filename with extension is passed separately from the directory structure on gcs. This replaces the `gs_url` param with `name` and `sub_bucket` (see 4723ab4)
- new functions `get_most_recent_from_bucket()` and `list_bucket_files()`to help identify the file on GCS with a name match, irrespective of if that file has a hash and version number.

Resolves #22, #24, #25

### DIFF
--- a/bedrock/utils/gcp.py
+++ b/bedrock/utils/gcp.py
@@ -22,6 +22,21 @@ GCS_CORNERSTONE = "gs://cornerstone-default"
 
 
 def download_gcs_file_if_not_exists(name: str, sub_bucket: str, pth: str) -> None:
+    """
+    Download a file from Google Cloud Storage (GCS) if it does not already exist
+    locally. This will also download any associated metadata files (e.g.,
+    _metadta.json) or log files.
+
+    Parameters
+    ----------
+    name : str
+        Target file name with extension, but without version or hash
+    sub_bucket : str
+        Subdirectory within the GCS bucket.
+    pth : str
+        Local file path where the file should be saved, including extension.
+    """
+
     os.makedirs(os.path.dirname(pth), exist_ok=True)
     if os.path.exists(pth):
         return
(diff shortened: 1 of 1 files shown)


bedrock/utils/gcp.py
```diff
@@ -0,0 +1,253 @@
...
+import google.cloud.storage
+import googleapiclient
+import googleapiclient.discovery
+import pandas as pd
+import tenacity
+from google.cloud.storage.blob import Blob
+
+logger = logging.getLogger(__name__)
+
+GCS_CEDA = "gs://cornerstone-default"
```
cclin130: we should rename this to GCS_CORNERSTONE

bedrock/utils/gcp.py
```diff
@@ -0,0 +1,253 @@
...
+GCS_CORNERSTONE = "gs://cornerstone-default"
+
+
+def download_gcs_file_if_not_exists(name: str, sub_bucket: str, pth: str) -> None:
+    os.makedirs(os.path.dirname(pth), exist_ok=True)
+    if os.path.exists(pth):
+        return
+    [
+        download_gcs_file(n, sub_bucket, pth)
+        for n in get_most_recent_from_bucket(name, sub_bucket)
```
cclin130: its confusing to me that `download_gcs_file_if_not_exists` has "file" as singular when it may actually return multiple files. If im understanding correctly, we loop to get the metadata and the parquet?

could you at least upload the docstring so that its clear it downloads the metadata as well? i think i follow what you're doing -- eventually we will have metadata files for CEDA parquets, and CEDA parquets will also be versioned with a hash?

cclin130: yep agreed, thank you for adding the docstring

PR #32, opened 2025-11-24
Rename `input_output` folder to `iot`

MoLi7:
Rename folder name to avoid confusion.

### DIFF
--- a/bedrock/ceda_usa/extract/iot/io_2017.py
+++ b/bedrock/ceda_usa/extract/iot/io_2017.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import os
-import posixpath
 
 import pandas as pd
 from typing_extensions import deprecated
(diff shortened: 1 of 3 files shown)


bedrock/ceda_usa/extract/iot/io_2017.py
```diff
@@ -137,8 +138,10 @@ def _load_2017_detail_make_use_usa(
     """
     df = (
         load_from_gcs(
-            name=USA_2017_DETAIL_IO_MATRIX_MAPPING[matrix_name],
-            sub_bucket=GCS_USA_DIR,
+            gs_url=posixpath.join(
```
bl-young: @MoLi7 I think there's a conflict here as I made some changes to these functions when i updated the extract logic. `gs_url` should not be a called parameter anymore

bl-young (approved):

PR #55, opened 2025-12-23
Refactor transform flowsa files

catherinebirney:
cc:

## What changed? Why?

- Move flowsa FBS related files into sub-directories within transform. The files within the subfolders must start with the name of the subfolder (ex. files within "GHG" must be named "GHG_") in order for functions to work correctly
- Move activity set csv files into relevant extract subdirectories by data source
- Update paths/locations where yaml loader looks for functions to generate the FBS

## Testing

- Locally generated 4 FBS with FBAs stored locally (including ghg m1/m2)
- Ran the GHG m1/m2 FBS where FBAs are downloaded from cornerstone google cloud storage
- Ran GHG m1/m2 FBS with FBS generated locally

### DIFF
--- a/bedrock/utils/config/flowsa_yaml.py
+++ b/bedrock/utils/config/flowsa_yaml.py
@@ -124,9 +124,7 @@ class FlowsaLoader(yaml.SafeLoader):
         # For security, this constructor does NOT search external config paths.
         # If someone who understands security concerns better than I do feels
         # it is safe to change this behavior, then go ahead.
-        module = importlib.import_module(
-            f'bedrock.transform.{module_name}'
-        )
+        module = importlib.import_module(f'bedrock.transform.{module_name}')
         return getattr(module, loader.construct_scalar(node))
 
 


bl-young:
Would there be any benefit to being consistent with case sensitivity in directory names? We are mixed (this applies to extract folder as well). On github, and I expect in some other cases, it impacts the sorting.

I assume dealing with this would be straightforward and would require one or two changes when looking for the folders.

catherinebirney:
@bl-young I addressed your comments

tested generating ghg m1/m2 with FBAs stored locally and with downloading FBAs

tested a few other FBS methods. They likely don't all run, but we can address down the road

bl-young (approved):

MoLi7:
@catherinebirney  some non-blocking q's:
- what does it mean for folder to have `_temp` as suffix?
- is testing the generation of FBAs within scope? asking b/c you tested generating ghg m1/m2 with FBAs stored locally and with downloading FBAs

PR #72, opened 2026-01-14
Remove ceda_usa folder

MoLi7:
cc:

## What changed? Why?

Removed the `ceda_usa` package from the `bedrock` directory, which should've been removed long time ago.

## Testing

confirming no one imports from `bedrock.ceda_usa`

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,13 +29,13 @@ After cloning the repository, in the root directory:
 4. **Confirm setup successful:**
 
     ```bash
-    uv run pytest ceda_usa/utils/__tests__/test_gcp.py
+    uv run pytest bedrock/utils/__tests__/test_gcp.py
     ```
 
-## Test CEDA matrix derivation
+## Test EEIO matrix derivation
 
 ```bash
-uv run pytest bedrock.transform/__tests__/test_usa.py -m eeio_integration
+uv run pytest bedrock/transform/__tests__/test_usa.py -m eeio_integration
 ```
 
 ## Outputs


.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         uses: ./.github/actions/setup-python
 
       - name: Typecheck w/ mypy
-        run: uv run mypy bedrock/ceda_usa
```
MoLi7: @cclin130 since I'm removing the entire `ceda_usa` folder, so here I changed it to `/extract/iot`. LMK if this looks wrong to you.

.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
...
+        run: uv run mypy bedrock
```
MoLi7: mypy on entire codebase

pyproject.toml
```diff
@@ -118,4 +118,49 @@ markers = [
...
+    'bedrock/utils/',
+    'bedrock/publish/',
+]
+
+[[tool.mypy.overrides]]
+module = [
+    "google.*",
+    "googleapiclient.*",
+]
+ignore_missing_imports = true
```
MoLi7: moved `.mypy.ini` here

.github/workflows/ci.yml
bl-young: yes agree. see other comment below

PR #78, opened 2026-01-21
Store and retrieve FBA and FBS from inside the package

bl-young:
cc:

## What changed? Why?

This updates how FBAs and FBSs are written and read to be placed inside the package in `extract/output_data` and `transform/output_data` respectively. New functions, `load_preprocessed_output()` and `write_fb_to_file()` are slightly modified versions of similar functions from esupy. This is slowly breaking the dependency on esupy.

`.gitignore` was modified to make sure the metadata json files do not get added (and also updated to reflect recent reorgs)

Additional work will be needed to evaluate the metadata generation itself to make sure it is still accurate. (e.g. in `metadata.getMetadata()`)

## Testing
Running the FBS test successfully generates and stores the FBAs locally and the FBS locally along with their metadata files
```
uv run pytest bedrock/transform/__tests__/test_fbs.py -m eeio_integration
```

image

image

### DIFF
--- a/bedrock/transform/flowby.py
+++ b/bedrock/transform/flowby.py
@@ -302,20 +302,20 @@ class _FlowBy(pd.DataFrame):
                 ]
             if attempt == 'generate':
                 flowby_generator()
-            df = load_preprocessed_output(file_metadata, str(pth))
-            if df is None:
-                log.info(
-                    f'{file_metadata.name_data} {file_metadata.category} '
-                    f'not found in {pth}'
-                )
-            else:
+            try:
+                df = load_preprocessed_output(file_metadata, str(pth))
                 log.info(
                     f'Successfully loaded {file_metadata.name_data} '
                     f'{file_metadata.category} from {output_path}'
                 )
                 break
+            except FileNotFoundError:
+                log.info(
+                    f'{file_metadata.name_data} {file_metadata.category} '
+                    f'not found in {pth}'
+                )
         else:
-            log.error(
+            raise FileNotFoundError(
                 f'{file_metadata.name_data} {file_metadata.category} '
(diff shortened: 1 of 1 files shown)


bedrock/utils/metadata/metadata.py
```diff
@@ -47,23 +49,37 @@ def set_fb_meta(name_data, category):
...
-    write_metadata_to_file(PATHS, fb_meta)
+    fb_meta.tool_meta = return_fb_meta_data(
+        source_name, config, fb_meta.category, **kwargs
+    )
+    fname = f'{fb_meta.name_data}_v{fb_meta.tool_version}'
+    if fb_meta.git_hash is not None:
+        fname = f'{fname}_{fb_meta.git_hash}'
+    fname = f'{pth}/{fname}_metadata.json'
+    with open(fname, 'w') as fi:
+        fi.write(json.dumps(fb_meta.__dict__, indent=4))
```
bl-young: this chunk of code is essentially copied in from `esupy.write_metadata_to_file()`

bedrock/utils/io/read.py
```diff
@@ -0,0 +1,51 @@
...
+            # Sort files in reverse order by ctime (creation time on Windows,
+            # last metadata modification time on Unix)
+            sorted_matches = sorted(
+                matches, key=lambda f: f.stat().st_ctime, reverse=True
+            )
+        # Return the path to the most recent matching file, or '' if no
+        # match exists.
+        if sorted_matches:
+            return f"{pth}/{sorted_matches[0].name}"
+    return None
```
MoLi7: I think it'd be more informative to return an error message specific to the function instead of None.

PR #83, opened 2026-01-22
Add API Key instructions for extract data

catherinebirney:
cc:

## What changed? Why?

- Updated `extract` README with instructions on creating `API_Keys.env` and listing which extract data sources might need an API key
- Updated links to bedrock sources
- Updated path to load `API_Keys.env`

## Testing

Generated extract data sources that call on APIs

### DIFF
--- a/bedrock/extract/README.md
+++ b/bedrock/extract/README.md
@@ -105,4 +105,19 @@ on. See the example below.
 
 ```
 
+## API Keys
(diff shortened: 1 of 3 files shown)


catherinebirney:
@jvendries Just a heads up that when this is pulled in you will need to move your API_Keys.env within `extract`

bl-young (approved):

Issue #101, opened 2026-01-24
Task 1.4: Fix type annotations in utils/metadata, logging, taxonomy (~24 errors)


MoLi7:
## Goal
Fix ~24 type errors in remaining utils submodules

## Files to Fix
- `bedrock/utils/metadata/metadata.py`
- Logging utilities
- Taxonomy mapping files

## Focus Areas
- Metadata dictionary types
- Logging function signatures
- Taxonomy crosswalk return types

## Validation
```bash
uv run mypy bedrock/utils/metadata/metadata.py bedrock/utils/logging/flowsa_log.py
```

## Dependencies
- None (can start immediately)
- Blocks: Task 1.6

PR #129, opened 2026-01-27
Fix type annotations in utils/metadata, logging

MoLi7:
cc:
Closes: #101

## What changed? Why?

This PR adds type hints to several functions in the logging and metadata modules to improve code clarity and enable better IDE support. It also includes several bug fixes:

1. In `flowsa_log.py`:
   - Added proper type annotations to function parameters and return types
   - Fixed a potential bug in the `format` method by adding a fallback to the INFO formatter when a level doesn't have a specific formatter

2. In `metadata.py`:
   - Added comprehensive type annotations to all functions
   - Fixed potential bugs in `process_nested_sources` by properly handling different types of input
   - Improved null checking in `get_source_metadata` to prevent AttributeError
   - Fixed type issues in `getMetadata` by ensuring proper string conversion for year parameter

## Testing

`uv run mypy bedrock/utils/metadata/metadata.py bedrock/utils/logging/flowsa_log.py`

### DIFF
--- a/bedrock/utils/logging/flowsa_log.py
+++ b/bedrock/utils/logging/flowsa_log.py
@@ -1,6 +1,7 @@
 import logging
 import shutil
 import sys
+from typing import Any
 
 from esupy.processed_data_mgmt import mkdir_if_missing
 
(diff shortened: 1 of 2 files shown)


bedrock/transform/__tests__/test_fbs.py
```diff
@@ -5,16 +5,6 @@
 from bedrock.utils.validation.validation import compare_FBS
 
 
-@pytest.mark.skip  # replaced by compare to remote test
-def test_generate_fbs() -> None:
```
catherinebirney: is it necessary to delete this? This test could be useful testing FBS methods when there aren't versions on remote

MoLi7: I think so!

Issue #80, opened 2026-01-21
evaluate FBA generation for USDA_CoA data


bl-young:
Required for GHG_national_m2

PR #132, opened 2026-01-28
Update GCS pull for USDA

bl-young:
cc: @catherinebirney 
Closes: #80 

## What changed? Why?

USDA_CoA_Cropland, USDA_CoA_Cropland_NAICS, and USDA_CoA_Livestock load csv files from GCS during FBA generation. To generate raw csv files during the API call, set `load_from_gcs` to False. 

csv files for USDA_CoA_Cropland and USDA_CoA_Cropland_NAICS added to GCS for 2022: `gs://cornerstone-default/ceda-usa/input/USDA_2022/*` USDA_CoA_Livestock not used in GHG method so not yet evaluated.

Some consolidation of the parsing functions was performed. USDA_CoA.py removed from the exceptions for linting.

## Testing

Adds new test `test_generate_fba_compare_to_remote()` to facilitate review of newly generated FBA files. 
USDA_CoA_Cropland_NAICS is identical, but very minor differences in USDA_CoA_Cropland.

### DIFF
--- a/bedrock/extract/usda/USDA_CoA.py
+++ b/bedrock/extract/usda/USDA_CoA.py
@@ -202,6 +202,7 @@ def coa_cropland_parse(
                 'ORGANIC STATUS',
                 'NAICS CLASSIFICATION',
                 'PRODUCERS',
+                'TYPOLOGY',
             ]
         )
     ]


bedrock/extract/usda/USDA_CoA.py
```diff
@@ -466,40 +411,7 @@ def coa_cropland_NAICS_parse(*, df_list, year, **_):
...
-            'zip_5',
-            'watershed_desc',
-            'region_desc',
-            'state_ansi',
-            'state_name',
-            'country_name',
-            'county_ansi',
-            'end_code',
-            'group_desc',
-        ]
```
MoLi7: are we not dropping these columns anymore? or they're handled differently now?

MoLi7:
@bl-young do we know why there are very minor differences in USDA_CoA_Cropland?

bedrock/extract/usda/USDA_CoA.py
```diff
@@ -643,13 +508,14 @@ def coa_common_parse(df: pd.DataFrame) -> pd.DataFrame:
     df['MeasureofSpread'] = "RSD"
     df['DataReliability'] = 5  # tmp
     df['DataCollection'] = 2
-
+    # Keep only necessary columns
+    df = df[flow_by_activity_fields]
```
bl-young: @MoLi7 here at the end we filter the whole df by just the necessary fields instead of removing columns by name throughout

bedrock/extract/usda/USDA_CoA.py
bl-young: see [here](https://github.com/cornerstone-data/bedrock/pull/132#discussion_r2734646861)

catherinebirney:
These code changes would not impact the results of the FBA generation, so the numbers were likely updated in NASS Quickstats. Unfortunately, USDA does not publish release dates or announce when they make changes to the available data. Since the changes in results are only for woody crops and sugarcane, it seems likely that the data was updated. 

The new data for corn and small grains are at the national level and replace 0 values, so those aren't updated/recalculated numbers. It looks like they added missing aggregated numbers. 

So, agree with @bl-young  - nothing to be concerned about

bl-young:
ooo and I found this: https://www.nass.usda.gov/Corrections/?log=QS

catherinebirney (approved):
Approving this PR as is - will address the changes to the USDA COA Cropland FBA separately, as those data differences are due to changes within the published source data, rather than caused by any code within this repo.

Documenting the issue in #136

PR #139, opened 2026-02-03
Add unit test for scrap adjustment math

MoLi7:
cc: 
Closes: #168 

## What changed? Why?

Added a unit test to validate the mathematical equivalence of two approaches for scrap correction in the EEIO derivation:
1. Using pandas' `divide` method along axis=0
2. Pre-multiplying by an inverse diagonal matrix

The test confirms that `Vnorm.divide(d, axis=0)` is equivalent to `np.diag(1/d) @ Vnorm`, which validates the correctness of the scrap correction formula used in the codebase.

## Testing

Created a test case with a 3x3 matrix and a divisor vector, then compared the results of both mathematical approaches to ensure they produce identical outputs.

### DIFF
--- a/bedrock/transform/__tests__/test_eeio_derivation.py
+++ b/bedrock/transform/__tests__/test_eeio_derivation.py
@@ -1,0 +1,60 @@
+"""Unit tests for the scrap correction in derived_2017.py."""
+
+import numpy as np
+import pandas as pd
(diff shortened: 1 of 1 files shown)


jvendries (approved):
Successfully ran test file locally 

Image

Issue #90, opened 2026-01-23
DIAG-002: Port compareOutputandLeontiefXDemand


MoLi7:
## Type
Diagnostic / Runtime Check

## Description
Validates that total output equals Leontief inverse times final demand: `q = L @ y` (or `x = L @ y` for industry models).

## R Source
[ValidateModel.R#L152-L180](https://github.com/cornerstone-data/useeior/blob/master/R/ValidateModel.R)

## Target Location
`bedrock/utils/validation/eeio_diagnostics.py`

## Acceptance Criteria
- [ ] Create function `diagnose_output_vs_leontief_demand(output, L, y, use_domestic=False, tolerance=0.01)` in `eeio_diagnostics.py`
- [ ] Support both domestic (`L_d`) and total (`L`) Leontief matrices
- [ ] Compute: `output_check = L @ y`
- [ ] Calculate relative difference: `rel_diff = (output - output_check) / output`
- [ ] Log warning if any sector exceeds tolerance
- [ ] Return `DiagnosticResult` with detailed sector-level differences
- [ ] Add docstring explaining the Leontief demand model

## Depends on
- UTIL-001 (uses the Ly calculation helper)

PR #150, opened 2026-02-04
Create validation and test functions for compare_output_vs_leontief_x_demand

jvendries:
cc:
Closes: #90

## What changed? Why?

Added a new validation function `compare_output_vs_leontief_x_demand` to verify that the total sector output matches the model calculation of L @ y (Leontief inverse multiplied by final demand). This validation ensures mathematical consistency in our economic input-output models.

The function:

- Compares total sector output (commodity or industry) against L @ y calculation
- Provides detailed diagnostics including failing sectors and relative differences
- Supports both total and domestic model variants

Also updated imports and exports in the `__init__.py` file to make the new function available through the validation module.

## Testing

Added a test case `test_compare_output_and_L_y` that:

- Tests the validation with both commodity and industry models
- Verifies domestic and non-domestic variants

### DIFF
--- a/bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
+++ b/bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
@@ -299,8 +299,10 @@ def test_compare_Uset_y_dom_and_q_usa() -> None:
     assert len(r_q_with_U_d_and_y_d_validation.failing_sectors) == 0
 
 
-# @pytest.mark.xfail(reason="Data manipulation for aligning with the CEDA schema. Need to resolve during method reconciliation.")
-# @pytest.mark.eeio_integration
+@pytest.mark.xfail(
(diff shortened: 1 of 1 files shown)


bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
```diff
@@ -293,3 +296,51 @@ def test_compare_Uset_y_dom_and_q_usa() -> None:
...
+)
+def test_compare_output_and_L_y(
+    modelType: str,
+    use_domestic: bool,
+) -> None:
+
+    # Load Aq model objects
+    Aq = derive_2017_Aq_usa()
+    A_d = Aq.Adom
+    A_imp = Aq.Aimp
```
MoLi7: @jvendries on variable naming convention, what do you think about `Adom` and `Aimp`? I prefer them over `A_d` and `A_i` since they are more informative.

MoLi7:
@jvendries thanks for digging into the errors, code change LGTM!

Issue #123, opened 2026-01-26
[CEDA Diagnostics] 4/8: Implement diagnostic helper functions


briantobin-99:
## Summary
Implement bedrock/utils/validation/ceda_diagnostics/helpers.py with core comparison functions, inflation adjustment, and data models.

### Pydantic Models
- [ ] CedaOldEfSet - holds raw and inflation-adjusted EF vectors
- [ ] CedaEfsForDiagnostics - holds all EF data for comparison

### Core Comparison Functions
- [ ] diff_and_perc_diff_two_vectors(vector_new, vector_old, reference_val_name)
- [ ] diff_and_perc_diff_two_sector_vectors() - simplified for US-only

### Summary Statistics
- [ ] calculate_summary_stats_for_ef_diff_dataframe()

### EF Comparison Builder
- [ ] construct_ef_diff_dataframe(ef_name, ef_new, ef_old_with_ma, ef_old_no_ma)

### Inflation Adjustment
- [ ] inflation_adjust_ef_denom_to_new_base_year() - uses bedrock inflation utilities

### Data Loader
- [ ] pull_ceda_efs_for_diagnostics() - returns CedaEfsForDiagnostics object

## Acceptance Criteria
- [ ] All comparison functions handle edge cases (div by zero, NaN)
- [ ] Inflation adjustment uses bedrock existing inflation utilities
- [ ] Pydantic models validate data correctly
- [ ] Unit tests for each helper function

PR #151, opened 2026-02-04
Adds helper functions for EF diagnostics

briantobin-99:
cc:
Closes: #123

## What changed? Why?

I added helper functions to perform the calculations necessary for the EF diagnostics. I also got rid of files we no longer need.

## Testing

I ran the functions on a simple example to make sure they worked.


MoLi7 (approved):
good start!

Issue #91, opened 2026-01-23
TEST-001: Port compareCommodityOutputXMarketShareandIndustryOutputwithCPITransformation


MoLi7:
## Type
Unit Test / CI Assertion

## Description
Tests that commodity output adjusted by CPI equals market share matrix times CPI-adjusted industry output. Validates price index consistency.

## R Source
[ValidateModel.R#L200-L240](https://github.com/cornerstone-data/useeior/blob/master/R/ValidateModel.R)

## Target Location
`bedrock/transform/__tests__/test_eeio_accounting.py`

## Acceptance Criteria
- [ ] Create new test file `bedrock/transform/__tests__/test_eeio_accounting.py`
- [ ] Add test `test_commodity_industry_output_cpi_consistency()`
- [ ] Load test data: `q`, `x`, `C_m` (market share matrix), CPI vectors
- [ ] Compute: `q_adj = q * cpi_commodity_ratio`
- [ ] Compute: `x_via_market_share = C_m @ (x * cpi_industry_ratio)`
- [ ] Assert: `np.allclose(q_adj, x_via_market_share, rtol=0.05)`
- [ ] Use bedrock's existing `inflate_to_target_year.py` utilities
- [ ] Add `@pytest.mark.eeio_integration` marker

PR #156, opened 2026-02-07
Create bedrock/transform/__tests__/test_eeio_accounting.py

jvendries:
cc:
Closes: #91

## What changed? Why?

Added a new unit test for EEIO accounting balance validations. This test verifies that commodity output adjusted by CPI equals market share matrix times CPI-adjusted industry output, which is a port of an existing [R validation](https://github.com/cornerstone-data/useeior/blob/master/R/ValidateModel.R#L141) from ValidateModel.R.

The test specifically:

- Derives 2017 make table (V), commodity output (q), and industry output (g) for the USA
- Computes the market share matrix
- Obtains inflation factors from reference data
- Creates commodity CPI by multiplying industry CPI with the market share matrix
- Calculates CPI ratios and checks that the adjusted outputs match within a specified tolerance

## Testing

The test is parameterized to run with base year 2017, target year 2023, and a tolerance of 0.01. It uses numpy's assert_allclose to verify that the CPI-adjusted commodity output matches the market share matrix multiplied by CPI-adjusted industry output within the specified tolerance.

### DIFF
--- a/bedrock/transform/__tests__/test_eeio_accounting.py
+++ b/bedrock/transform/__tests__/test_eeio_accounting.py
@@ -6,8 +6,6 @@ times CPI-adjusted industry output (ValidateModel.R#L200-L240).
 
 from __future__ import annotations
 
-import pytest
-
 from bedrock.transform.eeio.derived_2017 import (
     derive_2017_g_usa,
     derive_2017_q_usa,
@@ -18,7 +16,7 @@ from bedrock.utils.validation.eeio_diagnostics import (
 )
 
 
-@pytest.mark.eeio_integration
+# @pytest.mark.eeio_integration
 def test_commodity_industry_output_cpi_consistency(
     base_year: int = 2017,
     target_year: int = 2022,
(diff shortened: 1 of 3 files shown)


bedrock/transform/__tests__/test_eeio_accounting.py
```diff
@@ -0,0 +1,78 @@
...
+    [
+        (2017, 2023, 0.05),
+    ],
+)
+# @pytest.mark.eeio_integration
+def test_commodity_industry_output_cpi_consistency(
+    base_year: int,
+    target_year: int,
+    tolerance: float,
+) -> None:
```
bl-young: I would suggest that this function largely be a validation function, which then gets called as a test for a specific pair of years. I think that would be more similar to the other validation tests. That way, the test function itself wouldn't actually have any parameters, it would just be a call to

`validate_commodity_industry_output_cpi_consistency(2017, 2023, 0.05)`

Would that make it more consistent with your other PRs? 

Basically the general rule being that test function's don't have any novel code in them.

jvendries:
.

PR #186, opened 2026-02-11
Adds tab for diagnostics restricted to significant sectors

briantobin-99:
cc:
Closes:

## What changed? Why?

Adds tab to focus in on changes to significant sectors in the diagnostics process


MoLi7 (approved):
LGTM!

PR #198, opened 2026-02-13
Re-enable bedrock/extract/epa/ in project.toml after fixing type errors

MoLi7:
cc:
Closes: #196

## What changed? Why?

Fixed some more typing errors and removed the EPA directory from the mypy exclusion list in `pyproject.toml`.

### DIFF
--- a/bedrock/extract/epa/EPA_GHGI.py
+++ b/bedrock/extract/epa/EPA_GHGI.py
@@ -733,13 +733,8 @@ def ghg_parse(
                         df = df.drop(index)
                     else:
                         if apbe_value == 'N2O':
-                            df.loc[index, 'ActivityProducedBy'] = re.findall(  # type: ignore[index]
-                                r'\(.*?\)', apb_value
-                            )[
-                                0
-                            ][
-                                1:-1
-                            ]
+                            match = re.findall(r'\(.*?\)', apb_value)[0][1:-1]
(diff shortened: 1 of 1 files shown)


bedrock/extract/epa/EPA_GHGI.py
```diff
@@ -522,259 +522,263 @@ def ghg_parse(
...
                         apbe_value = apb_value.replace('Total ', '')
                         df = df.drop(index)
                     else:
                         if apbe_value == 'N2O':
-                            df.loc[index, 'ActivityProducedBy'] = re.findall(
+                            df.loc[index, 'ActivityProducedBy'] = re.findall(  # type: ignore[index]
                                 r'\(.*?\)', apb_value
-                            )[0][1:-1]
-                            df.loc[index, 'FlowName'] = 'N2O'
+                            )[
```
bl-young: ugh - this can't possibly be the right approach to formatting can it?

MoLi7: Good catch! This was wrong. Adding `# type: ignore[index]` to the end of the `re.findall(` line pushed it past the 88-char line limit, and then `ruff format` awkwardly broke the chained `)[0][1:-1]` across multiple lines.

It's fixed it by extracting the `re.findall(...)[0][1:-1]` into a match variable first, keeping the `.loc` assignment on a clean single line.

Issue #180, opened 2026-02-11
Implement B matrix transformation with price year adjustment


MoLi7:
**Workstream 4: IO Model Infrastructure**

> **Related Discussion:** [#16 - How do we transform environmental totals by industry into B matrix including price year adjustment?](https://github.com/cornerstone-data/methods/discussions/16)

Implement B matrix with price year adjustment.

**Files to create/modify:**
- Create: `bedrock/utils/config/methods/io_b_matrix_price_adjust.yaml` — method config for this choice
- Modify: `bedrock/transform/eeio/derived_2017.py` — add if-else gate for price-adjusted vs. unadjusted B matrix
- Modify: `bedrock/extract/iot/io_price_index.py` — gate price index integration
- Modify: `bedrock/transform/__tests__/` — tests for both gated paths

**Outcome:**
B matrix with price year adjustment. Gate OFF computes without adjustment.

**Verification:**
- [ ] `io_b_matrix_price_adjust.yaml` created with `enabled: true`
- [ ] If-else gating implemented in `derived_2017.py`
- [ ] Tests verify mathematical correctness for both paths

**Diagnostics:**
- [ ] In your PR, kick off a diagnostics run to generate results in Google Sheets
- [ ] Compare diagnostics outputs between baseline (gate OFF) and new method (gate ON)
- [ ] Flag and document any unexpected differences

PR #205, opened 2026-02-17
Convert gross industry output from before to after redefinitions

MoLi7:
cc: 
Closes: 

## What changed? Why?

Added functionality to derive gross industry output after BEA's redefinition adjustments. This serves as the foundation to address #180. Will integrate the after redef output in `derive_B` in a follow-up PR with proper config flag and if-else gating.

This implementation:

1. Created a new module `derived_gross_industry_output.py` with functions to:
   - Extract co-production entries from Make tables
   - Compute co-production ratios between industries
   - Apply redefinition adjustments to gross output vectors
   - Derive after-redefinition gross output for target years

2. Added a reference to the "before redefinition" Make table in matrix mappings and created a loader function `load_2017_V_before_redef_usa()` to access this data.

3. Added comprehensive unit and integration tests to verify the mathematical correctness of the implementation, including a round-trip test with 2017 benchmark data.

The implementation handles both unconstrained and constrained co-production ratios, with the latter providing more accurate adjustments by using the actual movement delta between before/after redefinition tables.

## Testing

Added a comprehensive test suite that includes:
- Unit tests for each component function
- Tests with small hand-calculated examples to verify the math
- An integration test that confirms the 2017 benchmark data round-trip works correctly
- Verification that column sum discrepancies match expected BEA rounding tolerances


jvendries (approved):
Successfully ran `test_2017_redefinition_roundtrip()` locally and inspected the `compute_coproduction_ratios` and `adjust_gross_output` to confirm the logic

PR #214, opened 2026-02-19
Add more metadata to diagnostics

MoLi7:
Added more metadata to diagnostics, including:

- git info, branch info, and PR info
- D and N summary stats for significant sectors


bl-young:
Have you run a test of this yet? Code looks fine but it'd be good to see a sample.

Issue #182, opened 2026-02-11
4.4 Test 3 options to derive A matrix


MoLi7:
We have 3 options, therefore there will be 3 PRs, each handles 1 option and corresponding diagnostics.
**Current approach** 
- **CEDA's approach** (unchanged from baseline), assuming there are meaningful changes in technology mix from 2017 to 2023: use existing 2017→2022 summary table adjustment, then extend 2022→2023 with inflation factors for stability. This should yield no change in EFs compared to baseline which is CEDA snapshot, because the snapshot will be inflation adjusted to match year info, so inflation effect will cancel out.

**Alternative approaches**
- **USEEIO's approach**: do nothing to 2017 detail A, assuming 2023 has the same technology mix with 2017
- **Scale A from 2017 to 2023 using only summary table information**
- **Scale A via price adjustment**

PR #228, opened 2026-02-24
Add config flags for A matrix derivation methods (Issue #182)

MoLi7:
## Summary

- Adds three boolean config flags to `USAConfig` for testing alternative approaches to derive the A matrix (Issue #182):
  - `scale_a_matrix_with_useeio_method` — do nothing; treat 2017 A as representative of target year
  - `scale_a_matrix_with_summary_tables` — scale 2017→target year using summary table ratios only
  - `scale_a_matrix_with_price_index` — inflate 2017→target year using price index only
- Each flag will be wired up in a separate stacked PR, one per method

## Test plan
- [ ] Stacked PRs add gating logic and YAML configs; each tested independently via the diagnostics GitHub Actions workflow

### DIFF
--- a/bedrock/utils/config/usa_config.py
+++ b/bedrock/utils/config/usa_config.py
@@ -40,7 +40,9 @@ class USAConfig(BaseModel):
     ### IO Methodology selection
     transform_b_matrix_with_useeio_method: bool = False  # DRI: mo.li
     implement_waste_disaggregation: bool = False  # DRI: jorge.vendries
-    # TODO: Add transform_a_matrix after we decide what to do
+    scale_a_matrix_with_useeio_method: bool = False  # DRI: mo.li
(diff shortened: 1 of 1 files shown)


bedrock/utils/config/usa_config.py
```diff
@@ -40,7 +40,9 @@ class USAConfig(BaseModel):
     ### IO Methodology selection
     transform_b_matrix_with_useeio_method: bool = False  # DRI: mo.li
     implement_waste_disaggregation: bool = False  # DRI: jorge.vendries
-    # TODO: Add transform_a_matrix after we decide what to do
+    scale_a_matrix_with_useeio_method: bool = False  # DRI: mo.li
```
WesIngwersen: All are set to False by default?

Issue #174, opened 2026-02-11
Track but not integrate Household emissions


MoLi7:
**Workstream 3: GHG Attribution Rules**

> **Related Discussion:** [#41 - Do we account for all emissions by households?](https://github.com/cornerstone-data/methods/discussions/41)

Track household emissions separately without integrating into IO model.

**Files to create/modify:**
- Create: `bedrock/utils/config/methods/ghg_household_tracking.yaml` — method config for this choice
- Modify: `bedrock/transform/ghg/GHG_national_m1_common.yaml` — add if-else gate for track-only vs. integrate

**Outcome:**
Household emissions tracked but separate. Gate OFF integrates them.

**Verification:**
- [ ] `ghg_household_tracking.yaml` created with `enabled: true`
- [ ] If-else gating implemented
- [ ] Household emissions excluded from industry coefficients when gate is ON
- [ ] Total tracked emissions reconcile with GHGI totals

**Diagnostics:**
- [ ] In your PR, kick off a diagnostics run to generate results in Google Sheets
- [ ] Compare diagnostics outputs between baseline (gate OFF) and new method (gate ON)
- [ ] Flag and document any unexpected differences

MoLi7:
This one won't use a custom config because we won't derive EFs for household.

catherinebirney:
The household emissions have been incorporated into the Cornerstone FBS - we did not have a specific PR for this task. The following activities map to F0100 (They might map to additional sectors as well)

1. Transportation Lubricants
2. Coal Residential
3. Fuel Oil Residential
4. Natural Gas Residential
5. Wood Residential
6. Total (All Fuels) Residential
7. Domestic Refrigeration
8. Residential Stationary Air Conditioning
9. Light-Duty Trucks - Households
10. Light-Duty Trucks
11. Motorcycles
12. Passenger Cars - Households
13. Passenger Cars
14. Recreational Boats

PR #240, opened 2026-03-02
GHG track emissions by final users

catherinebirney:
cc:
Closes:

Related to method choice #174 and discussion [#41](https://github.com/cornerstone-data/methods/discussions/41)

#248 attributes refrigerants to households

## What changed? Why?

Track emissions by final users -  attribute to `F01000`


bl-young:
I think we determined that this is basically handled in other PRs, each activity set at a time and doesn't require its own PR, is that correct @catherinebirney?

Issue #178, opened 2026-02-11
3.7 Hybrid BEA/NAICS Schema Implementation


MoLi7:
**Workstream 3: GHG Attribution Rules**

> **Related Discussion:** [#47 - Is the core schema for the GHG model based on BEA or NAICS schema?](https://github.com/cornerstone-data/methods/discussions/47)

Document and implement hybrid BEA/NAICS schema approach.

**Files to create/modify:**
- Create: `bedrock/utils/config/methods/io_bea_naics_schema.yaml` — method config for schema choice
- Create/Modify: `bedrock/utils/taxonomy/README.md` — document schema approach
- Modify: `bedrock/transform/ghg/GHG_national_m1_common.yaml` — add clarifying comments and if-else gate
- Modify: `bedrock/utils/taxonomy/bea_v2017_to_ceda_v7_helpers.py` — add docstrings

**Outcome:**
Schema approach documented and gated. Gate allows switching schema approaches.

**Verification:**
- [ ] `io_bea_naics_schema.yaml` created with `enabled: true`
- [ ] If-else gating implemented for schema selection
- [ ] Documentation complete for BEA vs. NAICS usage

**Diagnostics:**
- [ ] In your PR, kick off a diagnostics run to generate results in Google Sheets
- [ ] Compare diagnostics outputs between baseline (gate OFF) and new method (gate ON)
- [ ] Flag and document any unexpected differences

PR #249, opened 2026-03-04
Updates to use the cornerstone target schema which extends to all waste sectors

bl-young:
cc:
Closes: #178 

## What changed? Why?

Most of the hybrid schema was already implemented to develop the CEDA FBS method. This extends specifically to cornerstone to enable all waste sectors needed for disaggregation.

- run by setting `hybrid_bea_naics_schema_in_ghg_attribution` to True. See method `2025_usa_cornerstone_fbs_schema`

## Testing

[Diagnostics](https://docs.google.com/spreadsheets/d/1E7efpRSDO3Vkx4LRQSAtenlo2Sa5nQpsu0KOHlY5Gi4/edit?usp=sharing)

FBS for `GHG_national_Cornerstone_2023_update_schema` shows no net change in emissions but emissions attributed to 6 digit waste sectors not just 4 digit.


bl-young:
Diagnostics on this are not showing expected variability in the waste sectors:
image

@MoLi7  we might have to look closer at the implementation of the new schema from earlier in the diagnostics. Though, perhaps until the disaggregation process is complete maybe this is the expected outcome?

MoLi7 (approved):

PR #267, opened 2026-03-12
GHG: Align enteric fermentation and manure management with FLOWSA

MoLi7:
cc:
Closes:

## What changed? Why?

Switched manure management N2O from the CEDA approach (`EPA_GHGI_T_2_1` total → proportional attribution via `EPA_GHGI_T_5_7`) to the simpler FLOWSA approach (pull N2O directly from `EPA_GHGI_T_5_7` with direct attribution). Both paths produce the same total (16.8 N2O), but the FLOWSA path is simpler, avoids an extra allocation step, and includes animals that CEDA suppresses (American Bison, Mules and Asses).

Changes:
- New transitional FBS YAML `GHG_national_Cornerstone_2023_ag_livestock.yaml` overriding T_5_7 (removed `FlowName: ["CH4"]` filter) and disabling T_2_1 `manure_management_n2O` via `PASS`
- Updated `GHG_national_Cornerstone_2023.yaml` with the same changes directly
- Added config flag `update_enteric_fermentation_and_manure_management_ghg_method` in `usa_config.py` and wired it in `derived.py`
- New config `2025_usa_cornerstone_ghg_ag_livestock.yaml` to enable the method

## Testing

Run `diff_methods` comparing `GHG_national_CEDA_2023` vs `GHG_national_Cornerstone_2023_ag_livestock` and use `compare_FBS()` to verify N2O totals match. Use `run_method_diff.py` to generate before/after tables.

### DIFF
--- a/bedrock/utils/validation/run_method_diff.py
+++ b/bedrock/utils/validation/run_method_diff.py
@@ -1,0 +1,130 @@
+import numpy as np
+import pandas as pd
+
+from bedrock.transform.flowbysector import FlowBySector, getFlowBySector
+from bedrock.utils.config.schema import dq_fields
+from bedrock.utils.validation.validation import compare_FBS
+
+baseline = "GHG_national_CEDA_2023"
+test_method = "GHG_national_Cornerstone_2023_mobile_combustion"
+
+# Optional: show all data (baseline vs test side by side) for these MetaSources.
+# Leave empty to skip. Uses merge only — does not filter to differences.
+METASOURCES_TO_COMPARE: list[str] = [
+    "EPA_GHGI_T_3_13",
+    "EPA_GHGI_T_3_14",
+    "EPA_GHGI_T_3_15",
+    # "EPA_GHGI_T_3_45",
+    # "EPA_GHGI_T_3_47",
+    # "EPA_GHGI_T_3_49",
+]
+
+# Download and load from GCS (local directory needs to be empty of this
+# method to force new download)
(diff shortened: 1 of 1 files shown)


bl-young:
I was kind of expecting there to be some data differences here, but @catherinebirney may know better.

MoLi7:
@bl-young hey Ben — I had Claude dig into why there are no data differences, since you were expecting some. here's what it found:

### Raw T_5_7 N2O data

| Animal | FlowAmount (MMT CO2e) | Suppressed |
|--------|----------------------|------------|
| Beef Cattle | 6.2 | — |
| Dairy Cattle | 6.1 | — |
| Goats | **0.0** | **+** |
| Horses | 0.1 | — |
| Mules and Asses | **0.0** | **+** |
| Poultry | 2.3 | — |
| Sheep | 0.3 | — |
| Swine | 1.8 | — |
| **Sum** | **16.8** | |

T_2_1 "Manure Management" N2O total = **16.8** — exact match.

### Why no diff

The suppressed animals (Goats, Mules and Asses) have `FlowAmount = 0.0` in the parquet. American Bison doesn't even have an N2O row in T_5_7 (only CH4, also suppressed to 0.0). So while the FLOWSA path *includes* these animals, they contribute zero — making the two approaches numerically identical.

The "data loss" from the writeup is real in principle (suppressed = unknown, not truly zero), but as far as the stored data goes, both paths land on the same 16.8 total and same per-sector distribution.

Claude verified this by reading the raw `EPA_GHGI_T_5_7_2023` parquet directly, not just the FBS outputs.

PR #298, opened 2026-03-26
allow specifying fbs version and hash for download

catherinebirney:
cc:
Closes:

## What changed? Why?

Allows for specifying the version and hash of a file we want to download from GCS - no longer only downloads the most recent file, but does default to recent file if no version/hash included in filename.


catherinebirney:
@bl-young wanted to update this for use in the ghg FBS validation to GHGI values - now we can specify the exact FBS version to use

bl-young (approved):
minor typos

PR #302, opened 2026-04-02
Working Land National 2012 FBS

catherinebirney:
cc:
Closes:

## What changed? Why?
- Updating the Land National 2012 FBS to help test handling / approach for flows and LCIA methods in bedrock matrix transformation along with the GHG FBS
- Updated three activity names to align with the updated names in the 2024 MLU data release.
- Land related data added to Cornerstone GCS into folders: `BLM_PLS_2012`, `BLS_QCEW_2012`, `BLS_QCEW_2014` , `EIA_CBECS_Land_2012`, `EIA_MECS_Land_2014`, `USDA_2012`, and `USDA_MLU_2012`
- Updated the related above FBA yaml files to default downloading data from GCS rather than raw data extraction
- For now, can only generate this land FBS if use [mlu_updates](https://github.com/FLCAC-admin/fedelemflowlist/tree/mlu_updates) branch of fedelemflowlist

## Testing
Some differences for cropland between original flowsa output (fbs1) and bedrock output (fbs2). As cropland is proportionally attributed from "111" to 6-digit NAICS with USDA Census of Ag data, the change in allocation is likely due to updates within Census of Ag data. Summed Flow-By-Sector value for both flowsa and bedrock output is equal, at 8,916,448,180,762 m2.

image


pyproject.toml
```diff
@@ -29,6 +29,8 @@ dependencies = [
     "tenacity>=8.2.3,<9.0.0",
     "tqdm>=4.66.2,<5.0.0",
     "openpyxl>=3.1.2,<4.0.0",
+    "tabula-py==2.10.0",
+    "distro>=1.5.0,<2.0.0",
```
bl-young: what is distro used for?

bl-young (changes requested):
So far this is running fine for me to generate the FBAs and the underlying FBS (e.g. employment) in bedrock.

I think for each of these data sources, we need to add the `gcs_fxn` call and `extract_data_from_raw_sources: False` in order to force bedrock to pull the raw data from GCS. Like we do currently for EIA MECS Energy, GHGI, and CoA. This way it will always use the same raw data source until we push something new.

pyproject.toml
bl-young: Let me rephrase - so I expect distro is required for tabula, but shouldn't the requirements handle it just fine if we just add only tabula-py?

catherinebirney:
@bl-young updated FBA yamls used in Land FBS to default to downloading data from GCS. Dropped distro requirement, because as you stated, not necessary. Ready to be re-reviewed.

bl-young:
I'm having trouble with tabula and getting java installed, can you share the PLS FBA?

bl-young (approved):
I was able to successfully generate the full FBS

PR #316, opened 2026-04-15
Add 2019-2022 GHG Cornerstone FBS

catherinebirney:
cc:
Closes:

## What changed? Why?

Add new GHG Cornerstone FBS yamls for 2019-2022. These are based on the 2023 Cornerstone methodology, with table numbers updated based on data year. Some differences in allocation data set years (like USDA CoA Cropland). 

Specifically, the differences across files are:
- year anchor changed (2019/2020/2021/2022).
- coa_year changes (2017 for 2019–2020; 2022 for 2021–2023).
- stationary combustion table switches from EPA_GHGI_T_A_5 (2023) to year-specific T_A_6/7/8/9.
- non-energy fossil-fuel source uses EPA_GHGI_T_3_25b in new files (vs T_3_25 in 2023).

## Testing
Compared time series FBS results to GHGI Table 2.1
image

Almost identical to the validation check for the GHG m2 FBS that were generated in flowsa
image


bl-young (approved):
Note I added some details in the PR about the specific differences between methods to make it clear.

In subsequent PRs we will need to create model specs with these.

PR #334, opened 2026-04-27
USEEIO Baseline 2 - wire GHG FBS m2

bl-young:
cc:
Closes:
## What changed? Why?

This PR implements the GHG m2 path wiring for the USEEIO baseline comparison. Adds an output allocation for attributing NAICS to BEA sectors where the mappings are many:many (e.g. gov't, construction).

Staged scope in this PR:

- `load_E_from_flowsa` now supports explicit m2 method selection:
  - Adds `GHG_national_2023_m2` selection in `bedrock/transform/allocation/derived.py`
  - Refactors method selection into `_select_flowsa_ghg_method()`
- Replaces the old sector mapping helper name with a schema-neutral one:
  - `map_fbs_sectors_to_model_schema(...)`
  - Updated internal call site and related analysis doc reference
- Adds weighted NAICS->BEA mapping path for m2:
  - Uses `NAICS_to_BEA_Crosswalk_2017` as mapping basis
  - For waste NAICS (562*), applies a targeted override from `Cornerstone_2025` waste sector definitions (via helper)
  - Uses `derive_gross_output_after_redefinition(target_year=usa_ghg_data_year)` for weights
  - Enforces fail-fast behavior for one-to-many sectors with missing/zero output weights
  - Adds mass-conservation check on mapped `FlowAmount` totals (raises when tolerance exceeded)
- Tightens config invariants in `USAConfig`:
  - disallow `new_ghg_method=true` with `use_ghg_national_2023_m2=true` (existing)
  - add invariant: `use_ghg_national_2023_m2` requires `use_useeio_schema=true`
- Updates sensitivity config naming/semantics to match new invariant:
  - Renames `useeio_phoebe_23_restore_schema.yaml` to `useeio_phoebe_23_restore_schema_and_ghg.yaml`
  - Adjusts contents so schema restore no longer represents an invalid m2+non-USEEIO combination
- Extends config test coverage:
  - m2 requires USEEIO schema
  - m2/new_ghg mutual exclusion

Why this step:

- Ensures m2 is selected intentionally and mapped using an explicit NAICS->BEA crosswalk with output-based weighting.
- Protects existing paths by constraining new weighted logic to the m2 branch and enforcing invariant combinations at config validation time.
- Adds conservation/error checks to prevent silent mapping drift.

## Testing

Test script confirm succesful mapping of m2 to target schema.

### DIFF
--- a/bedrock/transform/allocation/derived.py
+++ b/bedrock/transform/allocation/derived.py
@@ -8,6 +8,9 @@ import pandas as pd
 from bedrock.transform.allocation.constants import EmissionsSource
 from bedrock.transform.allocation.registry import ALLOCATED_EMISSIONS_REGISTRY
 from bedrock.transform.flowbysector import FlowBySector, getFlowBySector
+from bedrock.transform.iot.derived_gross_industry_output import (
+    derive_gross_output_after_redefinition,
+)
 from bedrock.utils.config.common import load_crosswalk
 from bedrock.utils.config.usa_config import get_usa_config
 from bedrock.utils.emissions.ghg import GHG_MAPPING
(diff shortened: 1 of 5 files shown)


bedrock/transform/allocation/derived.py
```diff
@@ -25,6 +31,140 @@
...
+    mapping2.loc[~valid_weight, 'Allocation'] = 1.0 / group_size.loc[~valid_weight]
+
+    return mapping2[['Activity', 'Sector', 'Allocation']].reset_index(drop=True)
+
+
+def _should_use_output_weighted_mapping() -> bool:
+    return bool(get_usa_config().use_ghg_national_2023_m2)
+
+
+def _apply_cornerstone_waste_overrides(mapping: pd.DataFrame) -> pd.DataFrame:
```
WesIngwersen, line 108: Will this later need to modified for any disaggregated sectors?

WesIngwersen (approved):
Built useeioio_phoebe_23_restore_schema_and_ghg  successfully without error

bedrock/transform/allocation/derived.py
bl-young: yes, essentially we don't have a master NAICS -> Cornerstone crosswalk so the disaggregations need to be added separately. This can likely be streamlined once we cleanup our v0.2 models and configs.

PR #336, opened 2026-04-28
USEEIO Baseline 4 - implement B adjustment

bl-young:
cc:
Closes:

## What changed? Why?

This PR wires `use_useeio_B` into Cornerstone B derivation with the explicit USEEIO-style 3-step method:

1. Adjust target-year industry output to 2017 USD using chained-price index ratio.
2. Divide GHG totals by adjusted industry output to compute industry intensities.
3. Transform industry intensities to commodity intensities using market shares (`Vnorm`).

Before this change, `use_useeio_B` was validated in `USAConfig` but not used in the B calculation path. After this change:

- When `use_useeio_B: true`, `derive_cornerstone_B_via_vnorm()` uses the 3-step USEEIO-style method above.
- `derive_cornerstone_B_non_finetuned()` now routes `use_useeio_B: true` to `derive_cornerstone_B_via_vnorm()` (no B summary-scaling / inflation branch).
- Existing behavior remains unchanged when `use_useeio_B: false`:
  - if `use_E_data_year_for_x_in_B: true`, use `derive_cornerstone_B_via_vnorm()`
  - otherwise use scaled + inflated B path.

This makes the config flag operational and aligns PR4 with the plan intent to have an explicit USEEIO-style B method switch.

In the same PR stream, we also carried forward the USEEIO A-method parameter into the Phoebe config set:
`scale_a_matrix_with_useeio_method` is set to `true` across USEEIO Phoebe configs, with a dedicated
`useeio_phoebe_23_restore_useeio_A.yaml` sensitivity that flips only that parameter to `false`.

**Files touched**

- [`bedrock/transform/eeio/derived_cornerstone.py`](bedrock/transform/eeio/derived_cornerstone.py)
  - Added `use_useeio_B` branch in `derive_cornerstone_B_non_finetuned()`.
  - Added USEEIO-B price-adjusted-`x` logic in `derive_cornerstone_B_via_vnorm()`.
- [`bedrock/utils/economic/inflate_cornerstone_to_target_year.py`](bedrock/utils/economic/inflate_cornerstone_to_target_year.py)
  - Added `get_cornerstone_industry_price_ratio(...)` for industry chained-price adjustment used by USEEIO-B.
- [`bedrock/transform/eeio/__tests__/test_useeio_b_steps.py`](bedrock/transform/eeio/__tests__/test_useeio_b_steps.py)
  - Validates the 3-step USEEIO-B path (adjust x, divide, transform).

## Testing

- `python -m pytest bedrock/transform/eeio/__tests__/test_useeio_b_steps.py -q`

Diagnostics are running on full model `useeio_phoebe_23` and are looking good. Still some issues to review.


bl-young:
I'm reviewing a `D` matrix difference for coal mining (`212100`), where we currently show 2.47 while the D matrix value in Phoebe is 1.75. Relevant values out of bedrock:

```
sector_code | 212100
sector_name | Coal mining
E_total | 6.10E+10
x_nominal | 3.72E+10
industry_price_ratio_to_original_year | 0.6635
x_adjusted | 2.47E+10
Bi_total_industry | 2.47E+00
```

In useeior, I see the adjusted output (via [adjustOutputbyCPI](https://github.com/cornerstone-data/useeior/blob/90f4d7a022a3ecb2043535c30b473cd1068078a8/R/SatelliteFunctions.R#L83-L84)) for 212100 as 3.28E+10

Confirming `MultiYearIndustryOutput` as the following:
2023 output (2023 $): 4.94E+10
2017 output (2017 $): 3.69E+10

So, 3.28E+10 * 0.6635 = 
2023 output (2017 $): 3.28E+10

which yields a D value (unadjusted) of around 1.86. 

So what is driving this difference in industry output used between the two approaches?

bl-young:
For coal mining, there is a substantial difference in industry output in before redefinitions vs after redefinitions, so this further confirms I need to review this where it was flagged:
https://github.com/cornerstone-data/bedrock/blob/1523158e240b486220e21db63f6a026c70b9c096/bedrock/transform/allocation/derived.py#L74-L83

bl-young:
This was addressed in PR3 and diagnostics reveal that the coal EF is now identical.

bl-young:
Diagnostics for the full stack of PR's: [Diagnostics](https://docs.google.com/spreadsheets/d/1kOji1PqO-4VggJKYeR4NTh8CbrRpRwdiYjdWlXiQPIY/edit?usp=sharing)

bedrock/utils/config/configs/useeio_phoebe_23_restore_useeio_A.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+model_base_year: 2017
+iot_before_or_after_redefinition: before
+
+usa_io_data_year: 2017
+usa_ghg_data_year: 2023
+
+use_cornerstone_2026_model_schema: true
+load_E_from_flowsa: true
+use_E_data_year_for_x_in_B: false
```
MoLi7, line 12: this is default to `False`, can be removed
```suggestion
```

bedrock/utils/config/configs/useeio_phoebe_23_restore_useeio_A.yaml
```diff
@@ -0,0 +1,21 @@
...
+implement_waste_disaggregation: true
+
+# Restore A-scaling behavior to Anchor A for this single sensitivity.
+scale_a_matrix_with_useeio_method: false
```
MoLi7, line 16: same here
```suggestion
```

bedrock/utils/config/configs/useeio_phoebe_23_restore_useeio_A.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+use_useeio_schema: true
+use_useeio_B: true
+use_ghg_national_2023_m2: true
+skip_scrap_adjustment_in_vnorm: true
+new_ghg_method: false
```
MoLi7, line 22: and here
```suggestion
```

MoLi7:
@bl-young code change LGTM, but are you still investigating issues with coal and alike?

bl-young:
> @bl-young code change LGTM, but are you still investigating issues with coal and alike?

No the coal issue was addressed. It was due to the gross output had not yet been converted to before redefs. see https://github.com/cornerstone-data/bedrock/pull/336#issuecomment-4361350442

bl-young:
One of the largest remaining differences is `5416A0` (nearly double the D matrix value)
image

After some sleuthing have come to this conclusion:

The `V_n` column (from USEEIO) for `5416A0` almost matches Bedrock on the big diagonal (~90% supplied by `5416A0` itself), but differs mainly across disaggregated waste industries (`562*`), where shares are redistributed without changing the column sum much. When we keep Bedrock’s `E` and USEEIO-style `x` fixed and only swap that column to USEEIO’s shares, direct intensity D for `5416A0` drops to ~5.2×10⁻³, essentially lining up with the workbook’s ~5.35×10⁻³, versus ~9.6×10⁻³ with Bedrock’s `V_norm`. **So differences in embedded market shares—especially how waste co‑production is split—look like the main explanation for the diagnostic gap.**

bl-young:
@jvendries - remind me, the data you used for the waste disaggregation inputs is not exactly identical to what was used in useeio, is that correct? Since that was based on before redefinitions and now we have after redefinitions? So for a true replication we might actually need the original csv make and use data?

jvendries:
> @jvendries - remind me, the data you used for the waste disaggregation inputs is not exactly identical to what was used in useeio, is that correct? Since that was based on before redefinitions and now we have after redefinitions? So for a true replication we might actually need the original csv make and use data?

This is correct. useeior uses before redefinitions weights, while bedrock uses after redefintions. Both weights should be available in the respective repos.

WesIngwersen (approved):
Ran all configs and built all useeio phoebe models properly

bedrock/utils/config/configs/useeio_phoebe_23_restore_useeio_A.yaml
bl-young: I don't disagree - however for all these sensitivities I find it easier to leave them in and make it explicit so that its easier to look across all 5 or 6. So if ok by you (and since these aren't permanent configs) i'd prefer to leave them in.

bl-young:
> This is correct. useeior uses before redefinitions weights, while bedrock uses after redefintions. Both weights should be available in the respective repos.

Ok I will explore this tweak in a subsequent PR

PR #376, opened 2026-05-05
Update MECS allocation

catherinebirney:
cc: [#68](https://github.com/cornerstone-data/methods/discussions/68)
Closes:

## What changed? Why?

1. Update EIA MECS year to 2022 for Cornerstone GHG FBS 2021-2023 (align with COA data source years of 2017 and 2022)

1. When allocating industrial and manufacturing using EIA MECS, MECS relies on a GHGI table (Energy Consumption Data and CO2 Emissions from Fossil Fuel Combustion by Fuel Type) for further attribution. The CEDA approach uses the same data year for this GHGI table as for the overall GHG FBS method (So if the GHG FBS is for 2023, the MECS attribution uses the 2023 GHGI data). In this PR, we update the approach to match the GHGI year with the EIA MECS year. So we use the 2018 GHGI data to allocate the 2018 EIA MECS regardless of the year of the GHG FBS. 

    - By aligning the MECS and GHGI allocation years, fuel is no longer fully consumed by the manufacturing sectors, so our new method has additional rows for coal non-manufacturing sectors for tables 3.8, 3.9, and A.5

`GHG_national_Cornerstone_2023_v0.3.yaml` represents the primary Cornerstone FBS moving forward for future development, the current 2023 method `GHG_national_Cornerstone_2023` remains unchanged until we are ready for our next release.

`GHG_national_Cornerstone_2023_mecs.yaml` is the FBS method stub to use for testing atomic changes to the CEDA model.

## Testing

Comparing output to `GHG_national_CEDA_2023` with `GHG_national_Cornerstone_2023_mecs`
[Diagnostics](https://github.com/cornerstone-data/bedrock/actions/runs/26548684401)
[EF Diagostic Results](https://docs.google.com/spreadsheets/d/1d4pMGSf03rvox7h21Riy81lKsRcHN5BpA0iwxj1XDfY/edit?gid=149719632#gid=149719632)

Comparing impact on `GHG_national_Cornerstone_2023` results. Original (on `main`) and updated FBS have the same total flow amounts. 

Below are the comparisons for sector level FlowAmounts from FBS 1 (original output) to FBS 2 (updated method)
[GHG_2023_3digit_sector_comparison.csv](https://github.com/user-attachments/files/28362473/GHG_2023_3digit_sector_comparison.csv)
[All sector comparison](https://github.com/user-attachments/files/28362489/GHG_2023_comparison.csv)


bl-young:
Code changes look good. We need to think a bit about the strategy here about when to pull this in, and relatedly, what to run the diagnostics against. If we want to understand the impacts of this change alone, do we want to do an FBS method for 2023 with everything else the same (basically a copy of the FBS method before this change)? Then we can isolate the differences here.

catherinebirney:
> Code changes look good. We need to think a bit about the strategy here about when to pull this in, and relatedly, what to run the diagnostics against. If we want to understand the impacts of this change alone, do we want to do an FBS method for 2023 with everything else the same (basically a copy of the FBS method before this change)? Then we can isolate the differences here.

I updated the description above to include sector level comparisons from the original method to this new method

bl-young:
Any chance you can keep the activity sets in there? It'd be good to differentiate the different fuel types and confirm that the activity sets are moving in the direction we expect

bl-young:
We decided to treat this as a new change to evaluate w/ diagnostics so requires a new model config and flag.

catherinebirney:
@MoLi7 and @bl-young - The diagnostics are updated to reflect the new `GHG_national_Cornerstone_2023_mecs` FBS, comparing the impact of these changes to `GHG_national_CEDA_2023`. 

I'm going to generate new `GHG_national_Cornerstone_2023` results where we keep the activity names and compare to the current output on main, which, as Ben pointed out, should help us determine if the results make sense. I will tag again when those are added.

catherinebirney:
@MoLi7 and @bl-young I updated the csv attachments in this PR to include the metasources/activity sets. The results make sense to me - we have additional rows for non manufacturing sectors and the total FBS flowAmounts are equal.

bedrock/extract/epa/EPA_GHGI.py
```diff
@@ -803,7 +803,9 @@ def get_manufacturing_energy_ratios(parameter_dict: dict[str, Any]) -> dict[str,
         & (mecs['Unit'] == 'MJ')
     ].reset_index(drop=True)
 
-    # Load energy consumption data by fuel from GHGI
+    # Load energy consumption by fuel from GHGI - year is GHGI inventory year
+    # when defined in yaml (for older FBS), else MECS year.
+    # TODO: update year call to only use mecs_year when we drop the Cornerstone FBS flags
```
bl-young:
```suggestion
    # Load energy consumption by fuel from GHGI - year should
    # match MECS, but for older CEDA methods, the use of GHGI inventory year
    # is maintained when defined in yaml (for older FBS).
    # TODO: update year call to only use mecs_year (drop ghgi_year) when 
    # we drop the Cornerstone FBS flags
```

bl-young (commented):
Summarizing the change in results here for 2023. We use an estimate of fuel use from the GHGI compared to EIA MECS as a means to allocate industrial stationary combustion emissions between manufacturing (31-33) and non-manufacturing sectors. When we change the year for that comparison (from using the 2023 GHGI to the 2022/2018 GHGI) we see that there should be some coal consumption taking place in non-manufacturing sectors, hence we see an increase (addition of rows) for coal non-manufacturing. These emissions are taken proportionally from all manufacturing sectors 31-33.  For natural gas, we see the opposite impact, where emissions are shifted in aggregate from non-manufacturing to manufacturing.

The diagnostics (especially D diffs) appear to align with the changes we see in the FBS.

bl-young (approved):

PR #389, opened 2026-05-09
Do not call gcs unnecessarily when file to be loaded already exists locally

WesIngwersen:
This relates to the change to gcp.py occurring with this commit:
https://github.com/cornerstone-data/bedrock/commit/26136e7d72dc0fb6b344f835d490bf99ba6f271c#diff-30b839914f0f14c8e78090f9a82d33e3415d4e427a824916e5745440583fc0d7

The problem is the google cloud storage is being called when looking for any local data file in order to scan for metadata. 
Not only does this slow down the build process, it will cause fundamental model build code like 
`derive_Aq_usa()`  to fail if the user is offline or has issues with gcs authentication management. 

The fix moves the metadata-scanning block (including the __storage_client() call) inside the if not os.path.exists(pth): guard. Now the entire GCS interaction — both the main file download and the metadata scan — is skipped when pth already exists locally.

The tradeoff: if a new metadata file is added to GCS after the main file was already downloaded, it won't be picked up on subsequent runs. Given the function is named *_if_not_exists, that's consistent behavior — callers who need fresh metadata should delete the local file or use download_gcs_file directly.

The change to `build_model_manual_check.py` is inconsequential here and accidentally committed but just changed out the model being tested so it's more trouble to remove it than leave it in.

### DIFF
--- a/bedrock/utils/io/gcp.py
+++ b/bedrock/utils/io/gcp.py
@@ -93,23 +93,23 @@ def download_gcs_file_if_not_exists(name: str, sub_bucket: str, pth: str) -> Non
                 download_gcs_file(name, norm_bucket, pth)
             except Exception:
                 pass
-        # Download metadata from this GCS prefix when they exist
-        prefix = norm_bucket + "/"
-        gcs_bucket = __storage_client().bucket("cornerstone-default")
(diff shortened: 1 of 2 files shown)


catherinebirney (approved):
good call, Wes

PR #391, opened 2026-05-11
Add new validation function with case to check that if use_cornerstone_2026_schema then implement_waste_disaggregation

### LINKED ISSUE #390: Model can't use cornerstone_schema_2026 without implementing waste disaggregation
Although `use cornerstone_schema_2026` and  `implement_waste_disaggregation` are separate variables
https://github.com/cornerstone-data/bedrock/blob/7e6b8c8f29c8d74eef85dd111ad238ef09de34eb/bedrock/utils/config/usa_config.py#L56-L63

, if the former is true the latter must be true.  

To generate the model B, `derive_cornerstone_B_via_vnorm()` will be called
https://github.com/cornerstone-data/bedrock/blob/7e6b8c8f29c8d74eef85dd111ad238ef09de34eb/bedrock/transform/eeio/derived_cornerstone.py#L651-L662
which will in turn call `derive_E_usa()` and this series of functions

https://github.com/cornerstone-data/bedrock/blob/7e6b8c8f29c8d74eef85dd111ad238ef09de34eb/bedrock/transform/allocation/derived.py#L171-L198
ultimately in line 191 using the INDUSTRIES which are defined here and include the waste industries:

https://github.com/cornerstone-data/bedrock/blob/7e6b8c8f29c8d74eef85dd111ad238ef09de34eb/bedrock/utils/taxonomy/cornerstone/industries.py#L357-L366
with no gate to reaggregate/change with `implement_waste_disaggregation`

So the code needs to prevent these two from being cross.

WesIngwersen:
cc: @MoLi7 
Closes: #390 

## What changed? Why?
Adds new to validation function to usa_config.py 

## Testing
Add `test_disallow_not_implement_waste_disagg_when_cs_model_schema_flag()`  pytest.

### DIFF
--- a/bedrock/utils/config/__tests__/test_usa_config.py
+++ b/bedrock/utils/config/__tests__/test_usa_config.py
@@ -193,3 +193,17 @@ def test_disallow_useeio_b_with_e_data_year_x_flag() -> None:
             {'use_useeio_B': True, 'use_E_data_year_for_x_in_B': True},
             strict=True,
         )
+
+
+def test_disallow_not_implement_waste_disagg_when_cs_model_schema_flag() -> None:
+    with pytest.raises(
+        ValueError,
+        match='If use_cornerstone_2026_model_schema then implement_waste_disaggregation must be true',
+    ):
+        USAConfig.model_validate(
+            {
+                'use_cornerstone_2026_model_schema': True,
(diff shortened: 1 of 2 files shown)


WesIngwersen:
New test is currently failing.

PR #392, opened 2026-05-11
Cleanup diagnostics merge and add options for USEEIO baseline comparison merge

bl-young:
cc:
Closes:

## What changed? Why?

Cleans up `bedrock/utils/validation/analysis/` and makes `combine_ef_diagnostics` a consumer of diagnostics Google Sheets, and adds support for combining USEEIO Excel-baseline diagnostics runs.

### 1. File moves

- `bedrock/utils/validation/combine_ef_diagnostics.py` → `bedrock/utils/validation/analysis/` (and README entry added).
- `bedrock/utils/validation/run_method_diff.py` → `bedrock/utils/validation/fbs_compare/` (better folder home for an FBS-level diff script).

### 2. Refactor `combine_ef_diagnostics` to read Sheets directly

- Drops the assumption that inputs are downloaded `.xlsx` files; reads diagnostics tabs straight from a Drive folder of native Google Sheets via `bedrock.utils.io.gcp.`.
- Mirroring `diagnostics_plots` analyses, new `click` CLI: `--combo`, `--refresh`, `--output-xlsx`, `--output-sheet-id`. Local Excel workbook always written by default, to `analysis/output/<combo>/ef_diagnostics_merged.xlsx`. Google Sheets push is opt-in via `--output-sheet-id`.
- Output schema (`D_and_diffs_merged`, `N_and_diffs_merged`, `D_net_diff`, `N_net_diff`, `totals`, `totals_net_diff`, `config_summary_merged`) is identical to the legacy xlsx flow for combos that don't opt into new behavior, so the existing v0.2 destination Sheet is confirmed unchanged from the new CLI.

### 3. `combinations.py` registry

- Adds `ComboSpec` dataclass + `COMBINATIONS` dict. v0.2 is registered with its 17 input Sheet titles, target mapping, and Drive folder ID.
- `--output-sheet-id` is optional to push results to a Sheet.

### 4. USEEIO Excel-baseline mode + new `v0.3_useeio_phoebe` combo

Adds special handling when a set of diagnostics is run compared to USEEIO (`config_summary.diagnostics_baseline_source == 'gcs_useeio_xlsx'`):

- **Synthetic `pinned_useeio_baseline` column** (opt-in via `target_mapping`): when a combo references it, the merger injects a `pinned_useeio_baseline` column into `D_and_diffs_merged` / `N_and_diffs_merged` / `config_summary_merged`. Values come from the first input run's `D_old_inflated` / `N_old_inflated` (which IS the pinned USEEIO Excel baseline for USEEIO-comparison runs). The corresponding `config_summary_merged` column carries that run's pin metadata (`useeio_baseline_pin_*`, etc.). Combos that don't opt in (v0.2) keep their original output schema unchanged.
- **`totals` source picked per mode**:
  - Release-vs-snapshot combos: use the producer's pre-aggregated `BLy_and_E_orig_diffs` tab — unchanged from the legacy flow.
  - USEEIO-mode combos: the producer skips `BLy_and_E_orig_diffs` and instead uses `BLy_new_vs_BLy_old` across sectors.
- New `v0.3_useeio_phoebe` combo: Drive folder `1M2-Vopqfrx1vGcwoNi6wq55FmoELNV1s`, 7 input Sheet titles, target mapping where `useeio_phoebe_23` (the rebuild) targets `pinned_useeio_baseline` and every subsequent restoration step targets `useeio_phoebe_23` so the chain of divergence reads naturally.

### 5. Config rename: `useeio_phoebe_23_restore_useeio_{A,B}` → `..._cornerstone_{A,B}`

The two YAMLs actually *disable* the USEEIO-style flag (`use_useeio_B: false`, `scale_a_matrix_with_useeio_method: false`), i.e. they **restore the Cornerstone A/B approach**. Old name was backwards. 

## Testing

- `uv run python -m bedrock.utils.validation.analysis.combine_ef_diagnostics --combo v0.2`
  - Produces 7-tab workbook at `analysis/output/v0.2/ef_diagnostics_merged.xlsx` with the same schema as the legacy xlsx flow. 17 rows in `totals` / `totals_net_diff`, all per-config diffs match the prior destination Sheet.
- `uv run python -m bedrock.utils.validation.analysis.combine_ef_diagnostics --combo v0.3_useeio_phoebe`
  - Produces workbook at `analysis/output/v0.3_useeio_phoebe/ef_diagnostics_merged.xlsx` with the `pinned_useeio_baseline` column injected into D / N / config_summary tabs, totals sourced from `BLy_new_vs_BLy_old` aggregated to USA totals (7 rows).


bl-young:
Generates an output excel file to support the USEEIO baseline analysis which compiles data across the necessary diagnostics, e.g., "D_and_diffs_merged":
image

bl-young:
@WesIngwersen - just flagging for you that I changed two of the config names from `useeio_restore_useeio_A` to `useeio_restore_cornerstone_A` (and also for B), which I think makes the names clearer

MoLi7 (approved):
We reviewed this live together, LGTM!

PR #393, opened 2026-05-11
Add 2017 and 2018 GHG FBS

catherinebirney:
cc:
Closes:

## What changed? Why?

- Add FBS yamls for 2017 and 2018 GHG
- Note that we do not generate GHGI FBAs for these methods, the GHGI tables must be downloaded from GCS (with the extension v2.0.3)
- To generate these FBS, must set `download_sources_ok=True` when running `generateFlowBySector()`

## Testing

- compared sectors in 2017 and 2018 output to those in 2023


bl-young (approved):
I got tired of waiting for all the data suppression to be resolved, but it appeared to be downloading the v2.0.3 files just fine:

```
FlowBySector.generateFlowBySector("GHG_national_Cornerstone_2017", download_sources_ok=True)
```

PR #395, opened 2026-05-11
Add functionality to export waste-disaggregated matrices required as inputs for electricity disaggregation

jvendries:
cc: https://github.com/cornerstone-data/methods/issues/79
Closes:

## What changed? Why?

The waste disaggregation input CSVs (`WasteDisaggregationDetail2017_Use.csv` and `WasteDisaggregationDetail2017_Make.csv`) have been moved from `bedrock/extract/disaggregation/` into a dedicated `bedrock/extract/disaggregation/waste_disagg_inputs/` subdirectory. All references to these files across config files, source modules, and tests have been updated to reflect the new path.

A new `electricity_disagg_inputs/` subdirectory has also been added alongside `waste_disagg_inputs/`, with a `.gitkeep` to reserve the directory for future use.

A new `export_cornerstone_waste_disagg_matrices.py` module has been introduced to export waste-disaggregated Cornerstone matrices (V, Udom, Uimp, VA, full Y, and E) as CSVs into `electricity_disagg_inputs/`. This export validates that the active `USAConfig` matches the required methodology flags from `2025_usa_cornerstone_full_model.yaml` and that waste disaggregation weights are loadable before writing any files.

A new public function `derive_cornerstone_Ytot_full_cs_matrix()` has been added to `derived_cornerstone.py`, exposing the full commodity-by-final-demand Y matrix (including trade FD columns) as a copy to support the CSV export pipeline.

## Testing

New unit tests in `test_export_cornerstone_waste_disagg_matrices.py` cover precondition validation (passing, missing weights, and flag mismatch cases), CSV output correctness, and the copy behavior of `derive_cornerstone_Ytot_full_cs_matrix`. Existing integration tests in `test_waste_weights.py` and `test_waste_disaggregation.py` have been updated to point to the new file paths and continue to pass.


jvendries:
* **#395** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/395?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#411](https://github.com/cornerstone-data/bedrock/pull/411) <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/411?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#417](https://github.com/cornerstone-data/bedrock/pull/417) <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/417?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/395?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bl-young (changes requested):
If I'm reading this correctly, you need the matrices post-waste disaggregation as inputs to deriving the electricity disaggregation data?

What would you think about basing this off of the current snapshot instead of re-deriving? I think that would allow you to skip most of the code in `export_cornerstone_waste_disagg_matrices.py` and drop the associated tests and just pull the snapshot and save the csvs.

I'm a bit hesitant to store model outputs as csv inside the repo itself, but I'm not entirely clear on the next step in the process so I'm ok with that piece for now.

bl-young:
> What would you think about basing this off of the current snapshot instead of re-deriving? I think that would allow you to skip most of the code in `export_cornerstone_waste_disagg_matrices.py` and drop the associated tests and just pull the snapshot and save the csvs.
> 
Ok I realize now that not all the matrices you need are stored in the snapshots so that's not feasible. My primary concern is running the wrong function here so as not to get the right matrices, since our pipeline is not well established, which is why I had suggested pulling from the snapshots instead.

bedrock/extract/disaggregation/export_cornerstone_waste_disagg_matrices.py
```diff
@@ -0,0 +1,90 @@
...
+            "Cornerstone matrix export requires USAConfig methodology flags matching "
+            "`2025_usa_cornerstone_full_model.yaml`: "
+            + "; ".join(mismatches)
+            + ". Load that YAML via set_global_usa_config(...) before exporting."
+        )
+    if get_waste_disagg_weights() is None:
+        raise RuntimeError(
+            "Cornerstone matrix export requires waste disaggregation weights "
+            "(implement_waste_disaggregation and readable weight CSVs)."
+        )
```
bl-young, line 51: Could we simplify all this and just assert that the model config is the right one? `2025_cornerstone...` I don't think we want to re-establish the config specs in `_FULL_MODEL_MATRIX_EXPORT_FLAGS` as that just feels like more opportunity to become unlinked from the true config.

bedrock/extract/disaggregation/export_cornerstone_waste_disagg_matrices.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+    assert_cornerstone_matrix_export_preconditions()
+    out = (
+        _DISAGG_ROOT / "electricity_disagg_inputs"
+        if output_dir is None
+        else pathlib.Path(output_dir)
+    )
+    out.mkdir(parents=True, exist_ok=True)
+
+    V = derive_cornerstone_V()
```
bl-young, line 69: perhaps a note here to confirm that this follows the matrix derivation process from `xxx.py`

bl-young:
Looking at the data you need to export, I think all of them except for `E` should be covered by the matrices and objects in #406 (and #404). 

| Object | PR #395 (export path) | PR #406 (Excel publish) |
|--------|----------------------|-------------------------------|
| **V** | `derive_cornerstone_V()` | `_get_V` → `derive_cornerstone_V()` |
| **Udom** | `derive_cornerstone_U_set().Udom` | `_get_Udom` → `derive_cornerstone_U_set().Udom` (intermediate block; before `_assemble_extended_U`) |
| **Uimp** | `derive_cornerstone_U_set().Uimp` | Same `derive_cornerstone_U_set()` inside `_get_U`; not a separate sheet — recover as **Utotal − Udom** on commodity × industry |
| **Utotal** (`Udom + Uimp`) | Same components from `derive_cornerstone_U_set()` | `_get_U` → `uset.Udom + uset.Uimp` |
| **Y** (commodity × FD, cornerstone) | `derive_cornerstone_Ytot_full_cs_matrix()` *(public copy of full `Y`)* | `_get_U` → `_derive_cornerstone_Ytot_with_trade()[list(FINAL_DEMANDS)]` *(same pipeline; column order fixed to `FINAL_DEMANDS`)* |
| **VA** | `derive_cornerstone_VA()` | `_get_U` / `_get_Udom` → `derive_cornerstone_VA()` (stacked via `_assemble_extended_U`) |
| **E** | Exported in PR | Not produced by Excel publish on this branch |

bl-young:
Ignoring whether `E` is needed or not, I'd recommending either limiting this PR to just the reorganization of the disaggregation files without `export_cornerstone_waste_disagg_matrices.py` and its tests, and then just build your electricity disaggregation files off of the registry in #404 as needed in your next PR. Something like:

```
set_global_usa_config(CFG)

registry: list[SheetSpec] = _build_matrix_registry(CONFIG_NAME)

# name -> DataFrame | Series | None (None == sheet omitted in XLSX)
loaded: dict[str, pd.DataFrame | pd.Series | None] = {}
for spec in registry:
    loaded[spec.name] = spec.getter()

V = loaded['V']
U_extended = loaded['U']          # extended U (includes VA rows + FD cols on this branch)
U_d_extended = loaded['U_d']      # extended U_d (VA rows; FD cols truncated)
# ... same for 'A', 'B', 'q', etc.
```

Or don't use the registry and just replicate the chunk you have in `export_cornerstone_waste_disagg_matrices.py`
```
  V = derive_cornerstone_V()
  uset = derive_cornerstone_U_set()
  VA = derive_cornerstone_VA()
  Y = derive_cornerstone_Ytot_full_cs_matrix()
```
The latter might be easier. But ultimately if we have #406 you won't need to worry about writing them out as now we have a specific function for that.

jvendries:
> Ignoring whether `E` is needed or not, I'd recommending either limiting this PR to just the reorganization of the disaggregation files without `export_cornerstone_waste_disagg_matrices.py` and its tests, and then just build your electricity disaggregation files off of the registry in #404 as needed in your next PR. Something like:
> 
> ```
> set_global_usa_config(CFG)
> 
> registry: list[SheetSpec] = _build_matrix_registry(CONFIG_NAME)
> 
> # name -> DataFrame | Series | None (None == sheet omitted in XLSX)
> loaded: dict[str, pd.DataFrame | pd.Series | None] = {}
> for spec in registry:
>     loaded[spec.name] = spec.getter()
> 
> V = loaded['V']
> U_extended = loaded['U']          # extended U (includes VA rows + FD cols on this branch)
> U_d_extended = loaded['U_d']      # extended U_d (VA rows; FD cols truncated)
> # ... same for 'A', 'B', 'q', etc.
> ```
> 
> Or don't use the registry and just replicate the chunk you have in `export_cornerstone_waste_disagg_matrices.py`
> 
> ```
>   V = derive_cornerstone_V()
>   uset = derive_cornerstone_U_set()
>   VA = derive_cornerstone_VA()
>   Y = derive_cornerstone_Ytot_full_cs_matrix()
> ```
> 
> The latter might be easier. But ultimately if we have #406 you won't need to worry about writing them out as now we have a specific function for that.

Thanks @bl-young! This looks like a reasonable approach. However I'd like to clarify a couple of points:
1. I do think `E` will be needed for this PR; is that something you can add in #406 ?
2. Can you confirm that the "input matrices", that is, `V` and expanded `U` (which cotains `VA` and `Y` in it) are 
 A) **post-waste disaggregation**, meaning they have the waste disaggregated sectors with the waste-weights applied rather than the initial cornerstone duplicative expansion of `562000`, and  
 B) **have not** been year- or inflation-adjusted? 

These two items are important considerations in building the next steps of the electricity disaggregation (the next of which is electricity co-production reallocation). Specifically, the intent is to disaggregate based on the 2017 IO tables, and then apply the `A`/`B`/`x` transformations to the disaggregated tables.

bl-young:
> 2\. Can you confirm that the "input matrices", that is, `V` and expanded `U` (which cotains `VA` and `Y` in it) are
>     A) **post-waste disaggregation**, meaning they have the waste disaggregated sectors with the waste-weights applied rather than the initial cornerstone duplicative expansion of `562000`, and
>     B) **have not** been year- or inflation-adjusted?

Yes - that's the intention. These matrices should be the final matrices and they shouldn't be adjusted. But it would be good to confirm that, so let me know if you see otherwise.

jvendries:
> Ignoring whether `E` is needed or not, I'd recommending either limiting this PR to just the reorganization of the disaggregation files without `export_cornerstone_waste_disagg_matrices.py` and its tests, and then just build your electricity disaggregation files off of the registry in #404 as needed in your next PR. Something like...

@bl-young After spending some time iterating on the options for accomplishing this, I think it might be better to limit the interaction of the excel writer with this PR. The main reasons are: 
1) They are accomplishing different things: exporting useeior-style matrices for the excel writer vs. exporting cornerstone-ready consumable matrices for this PR; 
2) Getting the cornerstone-ready `U`-related matrices (`U.dom`, `U.imp`, `VA`, etc.) means either undoing the existing formatting and `U`-extension or adding a bunch of new `get*` functions to not use `U`-extension, which seems to defeat the purpose. 
3) Having `extract/disaggregation` use `publish/excel` objects inverts the normal `extract → transform → publish` dependency direction, which seems confusing. 

I will add another PR on top of this one to clean up the implementation of this PR and address your other [comment](https://github.com/cornerstone-data/bedrock/pull/395#discussion_r3223109926), but I think the interaction between this PR (and electricity disagg more generally) and the excel writer should remain minimal until/unless we want to publish the electricity disaggregated excel workbook.

I'm also adding a more thorough explanation of the above points in the collapsed section below. 



<details> 


<summary> Excel writer vs. this PR as source of matrices for electricity disaggregagtion </summary>
 
\
**Where to source the matrix values from**

Three options were considered:

- **(i) Import private getters by name** (`from bedrock.publish.excel.writer import _get_V, _get_Udom, ...`).
  - *Pros:* maximum reuse of the cached layer; aligns naively with "import as much as possible from publish".
  - *Cons:* (a) `_get_*` are all module-private (`_*` prefix) — relying on them is a fragile coupling; (b) `_get_U`/`_get_Udom` return *extended* shapes (intermediate + VA + optional FD) and apply `_apply_loc_suffix` `/US` decoration through `_materialize` — neither matches the PR1 CSV contract; (c) `_get_V` does match the PR1 V contract, but the rest of the PR1 outputs would still need direct calls to `derive_cornerstone_U_set / VA / Ytot_full / E` anyway, so this option provides almost no real reuse.
- **(ii) Reuse the exact `derive_*` calls publish makes internally** (`derive_cornerstone_V`, `derive_cornerstone_U_set`, `derive_cornerstone_VA`, `derive_cornerstone_Ytot_full_cs_matrix`, `derive_E_usa`).
  - *Pros:* zero dependence on publish's private surface; identical underlying logic to publish (publish itself calls these); preserves cornerstone shapes natively; works with the existing PR1 test scaffold (which monkeypatches these names on the exporter module).
  - *Cons:* none material — the exporter must still ensure cache coherence with publish if both run in the same process (mitigation: import and call `clear_publish_caches()` from publish, which IS the only public hook into the publish layer).
- **(iii) Promote `_get_*` to public in publish.** Note: simply renaming existing `_get_*` would not suffice — `_get_U`/`_get_Udom` return *extended* shapes (not the PR1 cornerstone shapes), `Uimp`/`VA` are only embedded inside the extended `U` block, and `E` has no getter at all. Delivering the PR1 contract via (iii) requires adding **five new public symbols** to `bedrock/publish/excel/writer.py` (`get_V`, `get_Udom_raw`, `get_Uimp_raw`, `get_VA`, `get_Ytot_full_cs`, `get_E`) plus deciding whether to register a new `E` sheet.
  - *Pros:* one genuine win — a process-level cache for `derive_E_usa()` (which is **not** `@functools.cache`d at source today, unlike `derive_cornerstone_V/U_set/VA/_Ytot_with_trade`). Cleaner public surface than (i). Could become a foundation for a future "publish/excel as canonical matrices API" if other consumers (diagnostics, supply-chain factors, comparison tooling) emerge.
  - *Cons:* (a) forbidden by C.1 (no modifications under `bedrock/publish/excel/`); (b) even with C.1 relaxed, layering is backwards — `extract/disaggregation/` consuming `publish/excel/` inverts the normal `extract → transform → publish` dependency direction, and "public" doesn't undo the inversion; (c) five of the six new getters would be one-line wrappers over already-`@functools.cache`d `derive_*` functions, so the cache value-add over (ii) is ≈ zero except for `E`; (d) the `E` cache benefit is independently and more cheaply obtainable by adding `@functools.cache` to `derive_E_usa()` directly in `bedrock/transform/allocation/derived.py` — no publish changes required; (e) expands publish's stable public surface for a single consumer that doesn't exist yet outside the exporter, growing maintenance/versioning surface in publish; (f) turns PR 1.2 into a multi-module change (extract + publish + publish tests).

**Recommendation:** Option (ii). Concretely, PR 1.2 reuses these symbols:

- From `bedrock.transform.eeio.derived_cornerstone`: `derive_cornerstone_V`, `derive_cornerstone_U_set`, `derive_cornerstone_VA`, `derive_cornerstone_Ytot_full_cs_matrix`.
- From `bedrock.transform.allocation.derived`: `derive_E_usa`.
- From `bedrock.publish.excel.writer`: **only** `clear_publish_caches` (public). Reason: if a caller has already run the publish pipeline in the same process, the publish getters will have cached results; the exporter should be able to invalidate those caches so a subsequent publish run sees fresh state (and vice versa). This is a one-line call, surfaces the soft dependency on publish in the exporter, and does **not** import any private getter.

Because publish/excel itself just delegates to the same `derive_cornerstone_*` calls, this choice does *not* fork the matrix materialization logic — it shares the underlying single-source-of-truth.

</details>

bl-young:
> I think the interaction between this PR (and electricity disagg more generally) and the excel writer should remain minimal until/unless we want to publish the electricity disaggregated excel workbook.

Ok that's fine

bedrock/extract/disaggregation/export_cornerstone_waste_disagg_matrices.py
jvendries: [Implemented](https://github.com/cornerstone-data/bedrock/blob/Electricity_disagg_PR_1_2/bedrock/extract/disaggregation/export_cornerstone_waste_disagg_matrices.py#L24-L31) in #417

bedrock/extract/disaggregation/export_cornerstone_waste_disagg_matrices.py
jvendries: Added a [note](https://github.com/cornerstone-data/bedrock/blob/Electricity_disagg_PR_1_2/bedrock/extract/disaggregation/export_cornerstone_waste_disagg_matrices.py#L64-L65) to this effect in #417

bl-young (approved):
Knowing that this gets further adjusted in #417 I am good with this.

jvendries:
### Merge activity

* **May 18, 8:28 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/395).
* **May 18, 8:29 PM UTC**: @jvendries merged this pull request with [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/395).

PR #396, opened 2026-05-11
Add GHG time series configs

catherinebirney:
cc:
Closes:

Will not pull this PR in - temporary/quick fix for time series. Will rework later. 

## What changed? Why?

Adds configs for 2017-2022 GHG FBS to enable generation of E matrix.

Temporary naming schema for the time series config files.


WesIngwersen (commented):
We already have `usa_ghg_data_year` which is used extensively in the model build process.
https://github.com/cornerstone-data/bedrock/blob/fe81865608fbd224884696f93ca71ef56d0936af/bedrock/utils/config/usa_config.py#L49
Recommend just expand the available year range on that variable

PR #404, opened 2026-05-12
prepares scaffolding for USEEIO style excel output for a model

bl-young:
cc: #401 
Closes:

## What changed? Why?

Stands up `bedrock/publish/` with the first publishable XLSX export of the bedrock EEIO model, mirroring the shape of `useeior::writeModeltoXLSX`.

### 1. New package layout

- `bedrock/publish/excel/writer.py` -- the workbook builder (`write_model_to_xlsx`) plus the `_build_matrix_registry`.
- `bedrock/publish/excel/cli.py` -- per-format Click CLI. Invoked as `uv run python -m bedrock.publish.excel.cli --config_name <cfg>`. Writes to `bedrock/publish/output/<git_sha>/<config_name>.xlsx`. Local only; GCS upload is TODO.
- `bedrock/utils/emissions/characterization.py` -- pure-math `build_ghg_characterization_matrix`.
- `bedrock/transform/eeio/derived.py` -- adds `derive_C_usa()` and `derive_D_usa()` wrappers (cached).

### 2. Sheet inventory

- **Real**: `B`, `C`, `D` (`= C @ B`), `q`, plus metadata (`commodities_meta`, `industries_meta`, `final_demand_meta`, `value_added_meta`, `config_summary`, `model_info`).
- **Placeholders** (registered with `getter = lambda: None`, sheets omitted via "skip if NULL" -- mirroring `useeior::WriteModel.R`): `V`, `U`, `U_d`, `A`, `A_d`, `A_m`, `L`, `L_d`, `M`, `M_d`, `M_m`, `N`, `N_d`, `N_m`, `Rho`, `Phi`, `Tau`, `x`, plus the metadata-side `demands`, `flows`, `indicators`, `SectorCrosswalk`. Each carries an inline TODO pointing at the missing derivation.

### 3. Known divergence from useeior (B units)

Bedrock `B` is in `kg CO2e / USD` (AR6 GWPs applied upstream at FBS aggregation in `bedrock/transform/allocation/derived.py`); useeior `B` is in physical `kg gas / USD` with a real GWP `C`. Consequences baked into the workbook:

1. The bedrock `B` sheet is NOT numerically comparable to useeior's `B` sheet. Row dimensions differ AND values differ per-row by a GWP factor.
2. The bedrock `C` sheet is a trivial `(1, 7)` row-summer (single `Greenhouse Gases` indicator, all ones), and `D = C @ B` reduces to `B.sum(axis=0)`. They are emitted for useeior sheet-name parity only.
[...]

Issue #401, opened 2026-05-12
publish model matrices to excel


bl-young:
Stand up a publishing pipeline under `bedrock/publish/`. Ultimately it should mimic the USEEIO model files (i.e., useeior's `writeModeltoXLSX`) even if not all matrices are available yet.

See also https://github.com/cornerstone-data/methods/issues/31

bl-young:
Core model matrices are complete so marking this as done. Later will come back to the final metadata matrices.

PR #406, opened 2026-05-12
update USEEIO style excel writer with more matrices

bl-young:
cc: #401 
Closes:

## What changed? Why?

Follows #404. Wires up the bulk of the useeior-style matrix and metadata set so the published workbook contains a usable EEIO model.

### 1. Matrix getters in `bedrock/publish/excel/writer.py`

Cached `_get_*` getters for V, U, U_d, A, A_d, B, C, D, L, L_d, M, M_d, N, N_d, q, x. Direct accessors wrap existing cornerstone derivations; L/M math comes from `bedrock.utils.math.formulas` (`compute_L_matrix`, `compute_M_matrix`); N is inline `C @ M`. U/U_d route through `_assemble_extended_U` -- see section 2.

Getters live in `writer.py` rather than `bedrock.transform.eeio.derived` because each has a single caller (the registry); promotion is appropriate if a second non-publish caller appears.

Supporting infrastructure:

- `_require_cornerstone()` -- single gate at the top of `write_model_to_xlsx`; raises `NotImplementedError` on legacy configs.
- `clear_publish_caches()` -- clears all `@functools.cache`-d getters in one call; invoked from `_helpers.py::clear_all_caches` for test isolation.
- Lazy imports of `bedrock.transform.eeio.derived(_cornerstone)` inside each getter to keep `import bedrock.publish.excel.writer` cheap.

### 2. Extended `U` matrix; truncated `U_d`

Useeior's `U` and `U_d` are four-block extended matrices:

```text
                 industries        FD codes
              +---------------+---------------+
commodities   | intermediate  |      Y        |
              +---------------+---------------+
VA codes      |     VA        |     0         |
              +---------------+---------------+
```

`_assemble_extended_U(intermediate, fd, va)` builds this layout. Both axes are tagged `Index.name = 'sector'` so `_apply_loc_suffix` adds `/US` to every label.

- **`U`** fully extended: intermediate = `Udom + Uimp`, FD = `_derive_cornerstone_Ytot_with_trade()[FINAL_DEMANDS]`, VA = `derive_cornerstone_VA()`, zero VA × FD corner.
- **`U_d`** has the VA rows but no FD columns. Bedrock has no `Ydom` matrix at FD-category resolution (only the `ydom` vector via `derive_cornerstone_ydom_and_yimp`), so the FD block is deferred. Shape: `(n_commod + n_VA) × n_ind`. Recorded in `model_info.u_d_extended` and `bedrock/publish/README.md`.

`_derive_cornerstone_Ytot_with_trade` is private; called from `writer.py` with an inline TODO to promote if a second non-publish caller appears.

### 3. Metadata sheets

- `flows` -- 7-row table over `bedrock.utils.emissions.ghg.GHG` (CO2, CH4, N2O, HFCs, PFCs, SF6, NF3). NOT comparable to useeior's ~2000-row elementary-flow list -- bedrock aggregates upstream at FBS aggregation.
- `indicators` -- 1-row table for `Greenhouse Gases`, matching `derive_C_usa()`.

### 4. Divergence-from-useeior callout (B, C, D, M, M_d, N, N_d)

- `B` is `kgCO2e / USD`; useeior `B` is physical `kg gas / USD`.
- `M = B @ L` and `M_d = B @ L_d` inherit B's units. NOT numerically comparable to useeior's M/M_d.
- `C` is a trivial `(1, 7)` row-summer. `D = C @ B`, `N = C @ M`, `N_d = C @ M_d` collapse to per-sector sums of B/M/M_d. Single `Greenhouse Gases` indicator; no GWP characterization.

Recorded in the writer module docstring, `model_info` (`divergence_from_useeior`, `u_extended`, `u_d_extended`), and `bedrock/publish/README.md`.

### 5. Sheet inventory

- **Matrices**: V, U, U_d, A, A_d, B, C, D, L, L_d, M, M_d, N, N_d, q, x.
- **Metadata**: flows, indicators, commodities_meta, industries_meta, final_demand_meta, value_added_meta, config_summary, model_info.
- **Omitted via "skip if NULL"** (each carries an inline TODO):
  - `A_m`, `M_m`, `N_m` -- blocked on real `B_imp`.
  - `Rho`, `Phi`, `Tau` -- no direct bedrock analogue.
  - `demands`, `SectorCrosswalk` -- pending schema design call.

### 6. Follow-ups

- `derive_cornerstone_Ydom_matrix()` -- closes the U_d FD-truncation divergence.
- Real `B_imp` -- unlocks A_m, M_m, N_m.
- Bedrock analogues for Rho, Phi, Tau.
- `demands` / `SectorCrosswalk` schema design.
- Resolve B-units divergence (paths in `bedrock/publish/README.md`).
- Promote `_derive_cornerstone_Ytot_with_trade` to public if another non-publish caller appears.

## Testing

- **Unit tests** (`bedrock/publish/__tests__/test_excel_writer.py`, 5 tests):
  - `test_assemble_extended_U_block_placement` -- four-block layout, zero VA × FD corner, `fd=None` branch.
  - Four registry-monkeypatching tests -- skip-if-None, empty-data-sheets guard, location suffix, round-trip.

  ```
  uv run pytest bedrock/publish/__tests__/test_excel_writer.py -v
  ```

- **End-to-end smoke** for `useeio_phoebe_23`: 24 sheets (16 data + 8 metadata) at `bedrock/publish/output/<git_sha>/useeio_phoebe_23.xlsx`. Wall time ~74s.

  ```
  uv run python -m bedrock.publish.excel.cli --config_name useeio_phoebe_23
  ```

- **Structural verification of U / U_d**:
  - U `(408, 425)` = `(405 commod + 3 VA) × (405 ind + 20 FD)`; U_d `(408, 405)`.
  - VA × FD corner zero; VA blocks byte-identical between U and U_d.
  - Intermediate `U - U_d >= 0` everywhere; ~19k cells differ.
  - VA codes (`V00100..V00300`) and FD codes (`F01000..F10S00`) match canonical phoebe-23 byte-for-byte.
  - Absolute counts diverge from canonical `(411, 428)` because bedrock cornerstone is 405-sector vs useeior 408 -- tracked separately.

### DIFF
--- a/bedrock/publish/excel/writer.py
+++ b/bedrock/publish/excel/writer.py
@@ -613,37 +613,6 @@ def _build_matrix_registry(config_name: str) -> list[SheetSpec]:
     ]
 
 
-# Sheet names treated as model "data" (matrices/vectors). At least one of
-# these must be non-None or `write_model_to_xlsx` raises -- otherwise we'd
-# silently publish a metadata-only workbook.
-_DATA_SHEETS: frozenset[str] = frozenset(
-    {
-        'V',
-        'U',
-        'U_d',
-        'A',
-        'A_d',
-        'A_m',
-        'B',
-        'C',
-        'D',
-        'L',
-        'L_d',
-        'M',
-        'M_d',
-        'M_m',
-        'N',
-        'N_d',
-        'N_m',
-        'Rho',
-        'Phi',
-        'Tau',
-        'q',
-        'x',
-    }
-)
-
-
 def _materialize(
     registry: Iterable[SheetSpec],
 ) -> list[tuple[str, pd.DataFrame | pd.Series]]:
@@ -677,9 +646,7 @@ def write_model_to_xlsx(out_path: str, *, config_name: str) -> None:
 
     Sheets are produced from `_build_matrix_registry(config_name)`. Any
     registry entry whose getter returns `None` is omitted from the
-    workbook (useeior-style "skip if NULL"). If zero data sheets
-    materialize, raises `RuntimeError` -- a metadata-only workbook is
-    treated as a silent failure, not a valid state.
+    workbook (useeior-style "skip if NULL").
 
     Raises `NotImplementedError` on legacy (non-cornerstone) configs --
     publish is wired only for cornerstone-schema configs today.
@@ -687,20 +654,10 @@ def write_model_to_xlsx(out_path: str, *, config_name: str) -> None:
     _require_cornerstone()
     registry = _build_matrix_registry(config_name)
     materialized = _materialize(registry)
-    data_count = sum(1 for name, _ in materialized if name in _DATA_SHEETS)
-    if data_count == 0:
-        raise RuntimeError(
-            'publish: no data sheets materialized; refusing to write a '
-            'metadata-only workbook. Check that derive_* getters in '
-            'MATRIX_REGISTRY are wired and that the requested config '
-            'supports them.'
-        )
     os.makedirs(os.path.dirname(out_path) or '.', exist_ok=True)
     logger.info(
-        'publish: writing %d sheets (%d data, %d metadata) to %s',
+        'publish: writing %d sheets to %s',
         len(materialized),
-        data_count,
-        len(materialized) - data_count,
         out_path,
     )
     with pd.ExcelWriter(out_path, engine='openpyxl') as writer:
--- a/bedrock/publish/excel/writer.py
+++ b/bedrock/publish/excel/writer.py
@@ -17,10 +17,14 @@ Practical consequences:
   * The bedrock `B` sheet is NOT numerically comparable to the useeior
     `B` sheet -- they have different row dimensions AND different
     per-row units.
-  * The bedrock `C` and `D` sheets are emitted for sheet-name parity
-    only. `C` is a trivial row-summer (single "Greenhouse Gases"
-    indicator over the 7 GHGs, all values = 1.0) and `D = C @ B`
-    reduces to `B.sum(axis=0)`. See
+  * `M` and `M_d` are `B @ L` / `B @ L_d`, so they inherit B's
+    `kgCO2e / USD` units. They are NOT numerically comparable to
+    useeior's physical-mass `M` / `M_d` either.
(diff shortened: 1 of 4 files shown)


bedrock/publish/excel/writer.py
jvendries: Consider if it would be worth it to add a new field like `kind` to the [SheetSpec class](https://github.com/cornerstone-data/bedrock/pull/406/changes#diff-0c7e367f07d3153c5d68273e822f4a60cfa19fab5d8b08d7776d4581d4880b46R84-R86) and tag each sheet either `data` or `metadata` so that `_DATA_SHEETS` can be removed? So for example: 

```
def _build_matrix_registry(config_name: str) -> list[SheetSpec]:
    return [
        # --- useeior matrices (order matches `matrices` in WriteModel.R) ---
        SheetSpec('V', _get_V, kind='data'),
        SheetSpec('U', _get_U, kind='data'),
...
 SheetSpec('demands', lambda: None, kind='metadata'),
        SheetSpec('flows', lambda: _build_flows_df(), kind='metadata'),
...
```
The idea being to remove the need to manually update `registry` and `datasheets` separately any time there is an update.

jvendries (changes requested):
Overall looks good. Ran the `bedrock/publish/__tests__/test_excel_writer.py` tests successfuly. 

However, I would request that the `E` matrix be included, or explicitly call out its exclusion. This is so that matrix is handled correctly in #395

bl-young:
> Overall looks good. Ran the `bedrock/publish/__tests__/test_excel_writer.py` tests successfuly.
> 
> However, I would request that the `E` matrix be included, or explicitly call out its exclusion. This is so that matrix is handled correctly in #395

Yes, given that `E` is not in the traditional excel output, I'm hesitant to write it here as part of this pipeline. Do you think that's something you could add on your end? I'm also open to thinking about adding it here in a subsequent PR.

bedrock/publish/excel/writer.py
bl-young: I agree with that concern of having to track those both together. Honestly I think this extra check is probably not necessary at all so i removed the whole thing.

jvendries (approved):

jvendries:
Changes look good.I tested locally without errors, and I'll add the `E` matrix "export" in the electricity disagg PR #395

PR #412, opened 2026-05-15
use only 1 inflation config flag apply_inflation_to_V. 

### LINKED ISSUE #410: V not getting inflated for q calculation even when apply_inflation_to_V = True
`apply_inflation_to_V:True`  applied inflation to V before getting a commodity_price_ratio in `get_vnorm_adjusted_commodity_price_ratio()`

https://github.com/cornerstone-data/bedrock/blob/dae2841ef0ef79f596059589f6b7e2c5135e8367/bedrock/transform/eeio/derived_cornerstone.py#L284-L326

 but it is not applied in `derive_cornerstone_V()`

https://github.com/cornerstone-data/bedrock/blob/dae2841ef0ef79f596059589f6b7e2c5135e8367/bedrock/transform/eeio/derived_cornerstone.py#L193-L204

which is called through `derive_cornerstone_Aq()` to get `q`:

https://github.com/cornerstone-data/bedrock/blob/dae2841ef0ef79f596059589f6b7e2c5135e8367/bedrock/transform/eeio/derived_cornerstone.py#L485-L492

I propose to simply move the if  `apply_inflation` section from derive_cornerstone_Vnorm_scrap_corrected() to inside of derive_cornerstone_V() so its used earlier in the pipeline.

WesIngwersen:
Deflate V in all cases to model_base_year when apply_inflation_to_V: True

Previously V was only being inflated when performing commodity-based price adjustment. 

This change will consistently apply V inflation to derive q as well. 

Replace update_inflation_factors flag which was only  used in 1 place.


WesIngwersen:
This is in draft because this function, where `update_inflation_factors` was used before and in draft is replaced, is a very confusing use of this param and needs to be addressed concurrently so as not to further confuse things

https://github.com/cornerstone-data/bedrock/blob/dae2841ef0ef79f596059589f6b7e2c5135e8367/bedrock/utils/economic/inflation_helpers_cornerstone.py#L52-L85

WesIngwersen:
> This is in draft because this function, where `update_inflation_factors` was used before and in draft is replaced, is a very confusing use of this param and needs to be addressed concurrently so as not to further confuse things
> 
> https://github.com/cornerstone-data/bedrock/blob/dae2841ef0ef79f596059589f6b7e2c5135e8367/bedrock/utils/economic/inflation_helpers_cornerstone.py#L52-L85

I added these back for now due to more thought needed to tackle this separate config flag.

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -190,7 +190,9 @@ def get_waste_disagg_weights() -> DisaggWeights | None:
 
 @functools.cache
 @pa.check_output(CornerstoneVMatrix.to_schema())
-def derive_cornerstone_V() -> pd.DataFrame:
+def derive_cornerstone_V(
+    apply_inflation: bool = False, target_year: int = 0
+) -> pd.DataFrame:
```
WesIngwersen: A question for @jvendries @bl-young @MoLi7 .  There is a mix in the code of calling this with params and also calling this without params and just using the usa_config vars() inside the function.  Currently its defaulting to not inflate, but I think we are moving toward a default of inflate based on the model year.  What if the defaults in the function call are   
```
def derive_cornerstone_V(
    apply_inflation: bool = get_usa_config().apply_inflation_to_V, target_year: int = get_usa_config().model_year
} -> pd.DataFrame:
```

I suggest that because the params would need to change in a time series is used for a model, unless that model_config var `model_year` dynamically changes within the time series preparation.

bl-young: I could be wrong, but I think calling `get_usa_config()....` inside the function definition is risky because that gets called when the function is imported and not when its used. E.g. see [here](https://medium.com/@nebiyuelias1/be-careful-when-using-default-arguments-in-python-fd92df94efee).

We could set `apply_inflation` to a `bool | None` such that by default it uses the config, e.g.,
```
if apply_inflation is None:
   use the config to set the value
if apply_inflation:
   apply inflation function
else:
  nothing

WesIngwersen:
@jvendries or @bl-young you will need to approve this in order to unblock the merge. thanks

bl-young (approved):

PR #413, opened 2026-05-15
replace `use_useeio_B` with `deflate_x_to_detail_io_year_for_B`

bl-young:
cc:
Closes:

## What changed? Why?

- Renames `use_useeio_B` to `deflate_x_to_detail_io_year_for_B` which better describes its function and clarifies that it is not mutulaly exclusive (but rather follows from) `use_E_data_year_for_x_in_B`. 
- Deflation now requires `use_E_data_year_for_x_in_B` at config load.
- Unused `bea_B` / `bea_E` helpers are removed from `cornerstone_bea_intermediates.py`. 
- USEEIO Phoebe YAMLs use the new flag name.

## Testing

Pipeline not changed. Tests pass.

### DIFF
--- a/bedrock/transform/eeio/__tests__/test_useeio_b_steps.py
+++ b/bedrock/transform/eeio/__tests__/test_useeio_b_steps.py
@@ -39,8 +39,8 @@ def test_useeio_b_adjust_divide_transform_steps(
         dc,
         "get_usa_config",
         lambda: SimpleNamespace(
-            use_useeio_B=True,
-            use_E_data_year_for_x_in_B=False,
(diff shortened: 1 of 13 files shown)


bl-young:
@WesIngwersen - any use of `use_useeio_b` will need to be updated in your script depending on the order in which we pull them in. If your branch gets pulled in first I can rebase here and make that change or vice versa.

WesIngwersen (approved):

bl-young:
@WesIngwersen do you want me to go ahead and pull this in or wait on anything on your branch?

PR #414, opened 2026-05-15
Add real vs summary derived q plots

WesIngwersen:
The summary approach creates an estimated q which can be compared to a more "real" q  or Cornerstone's best approximation of q. It generates a new plot of q diffs by year and summary deviation statistics.   summary_a_errors.py is expanded and renamed summary_a_q_errors.csv.


WesIngwersen:
@MoLi7 I believe the conflicts to be resolved are in some of your work that this branch was not updated with. I would be more comfortable if you resolved these to make sure we keep the changes in that compare_method_stability file that you want to keep.

WesIngwersen:
> @MoLi7 I believe the conflicts to be resolved are in some of your work that this branch was not updated with. I would be more comfortable if you resolved these to make sure we keep the changes in that compare_method_stability file that you want to keep.

I went ahead and accepted all incoming changes to merge and remove conflicts since this was stalled out here.

MoLi7 (approved):

Issue #384, opened 2026-05-07
EF inflation adjustment helper is not using model price ratios


WesIngwersen:
https://github.com/cornerstone-data/bedrock/blob/7e6b8c8f29c8d74eef85dd111ad238ef09de34eb/bedrock/utils/validation/diagnostics_helpers.py#L434-L463

Effects the validity of [N time-series results](https://github.com/cornerstone-data/bedrock/pull/381). 

This function goes back to calculating original price ratios from reference data which is problematic at multiple levels.  There are schema issues as this function to return the reference data is tied back to CEDA sector lists.  And f the model is using commodity-based inflation factors then in-principle they would be used here as well.

PR #415, opened 2026-05-15
inflate factors in diagnostics for USEEIO path

bl-young:
cc:
Closes:

When USEEIO-style configs use `deflate_x_to_detail_io_year_for_B`, new `D` and `N` are in `usa_detail_original_year` chain dollars while the baseline side is already inflated to `model_base_year`. EF diagnostics add optional `D_new_inflated` and `N_new_inflated` via the same sector price-index adjustment as `D_old_inflated` / `N_old_inflated` (`inflation_adjust_ef_denom_to_new_base_year`).

- `pull_efs_for_diagnostics` emits inflated vectors when `d_n_new_inflated_eligibility` allows it (skips identity year, double-apply, and legacy paths).
- `calculate_ef_diagnostics` inserts `{D|N}_new_inflated` before `{D|N}_new` and recomputes `{D|N}_perc_diff` against `{D|N}_old_inflated`.
- `combine_ef_diagnostics` prefers `D_new_inflated` / `N_new_inflated` when merging runs.

Does **not** address [#384](https://github.com/cornerstone-data/bedrock/issues/384) (EF inflation helper still uses sector reference-data price ratios via `obtain_inflation_factors_from_reference_data`, not model commodity/V-norm ratios). This PR reuses that helper for `D_new_inflated` / `N_new_inflated` so new and old sides stay comparable in diagnostics; fixing #384 is separate work.

## Testing

```powershell
uv run pytest bedrock/utils/validation/__tests__/test_diagnostics_helpers.py bedrock/utils/validation/__tests__/test_calculate_ef_diagnostics.py bedrock/utils/validation/__tests__/test_diagnostics_sector_alignment.py -q
```

Results in [diagnostics like this one](https://docs.google.com/spreadsheets/d/1WeabDjR9-y9rtfGvGmTl8dp2Ch-o7miNIvsMBdpr4wI/edit?usp=sharing)

### DIFF
--- a/bedrock/utils/validation/diagnostics_helpers.py
+++ b/bedrock/utils/validation/diagnostics_helpers.py
@@ -5,7 +5,8 @@ This module provides:
 - Core comparison functions (diff, percent diff)
 - Summary statistics calculations
 - Inflation adjustment for EF denominators
-- Data loading for diagnostics
+- Data loading for diagnostics (optional ``D_new_inflated`` / ``N_new_inflated`` when
+  eligible; merged into ``D_and_diffs`` / ``N_and_diffs`` by ``calculate_ef_diagnostics``)
 - Schema alignment for CEDA v7 ↔ cornerstone diagnostics
 """
 
@@ -19,7 +20,7 @@ import numpy as np
 import pandas as pd
 from pydantic import BaseModel
 
-from bedrock.utils.config.usa_config import get_usa_config
+from bedrock.utils.config.usa_config import USAConfig, get_usa_config
 from bedrock.utils.economic.inflation_helpers_ceda import (
     obtain_inflation_factors_from_reference_data,
 )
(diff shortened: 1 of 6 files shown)


bedrock/utils/validation/diagnostics_helpers.py
```diff
@@ -65,17 +66,69 @@ class EfsForDiagnostics(BaseModel):
...
     D_old: OldEfSet
     N_old: OldEfSet
+    D_new_inflated: ta.Optional[pd.DataFrame] = None
+    N_new_inflated: ta.Optional[pd.DataFrame] = None
 
     class Config:
         arbitrary_types_allowed = True
 
 
+def d_n_new_inflated_eligibility(cfg: USAConfig) -> tuple[bool, str]:
```
MoLi7, line 88: why do we need this?

bl-young: its just a config checker (with explanation) to make sure that the `{D|N}_new_inflated` is only run when appropriate

MoLi7 (approved):
d_n_new_inflated_eligibility seems handy as a guardrail for us

PR #411, opened 2026-05-14
"Export of waste-disagg matrices through excel publish functionality"

jvendries:
cc:
Closes:

## What changed? Why?

Introduces `electricity_disagg_cornerstone_materializer.py` as a single-sourced module that materializes the six cornerstone frames (`V`, `Udom`, `Uimp`, `VA`, `Y`, `E`) required for electricity disaggregation CSV export. This consolidates what was previously scattered `derive_*` calls in both `export_cornerstone_waste_disagg_matrices.py` and `writer.py` into one canonical location.

`_get_V()`, `_get_U()`, and `_get_Udom()` in `writer.py` now delegate to `materialize_electricity_disagg_cornerstone_frames()` and `intermediate_use_totals()` from the new materializer, ensuring the writer and the CSV exporter are always in sync.

`export_cornerstone_waste_disagg_matrices.py` is updated to read expected methodology flags directly from the pinned YAML (`2025_usa_cornerstone_full_model.yaml`) via `USAConfig.model_validate` rather than a hardcoded dict, so the precondition check stays aligned with the YAML automatically. A new public entry point `export_electricity_disagg_cornerstone_csvs` is added (with `export_cornerstone_matrices_to_csv` kept as a backward-compatible alias) that clears publish caches before materializing.

## Testing

Three new smoke tests cover the materializer:
- `test_materialize_electricity_disagg_cornerstone_frames_smoke` — asserts all six frames are returned with consistent shapes and aligned indices.
- `test_intermediate_use_totals_matches_u_dom_plus_u_imp` — asserts `intermediate_use_totals()` equals `Udom + Uimp` from the full frame bundle.
- `test_writer_getters_match_materializer_v_and_u_intermediate` — asserts `_get_V()`, `_get_U()`, and `_get_Udom()` from the writer return values consistent with the materializer output.

Existing tests in `test_export_cornerstone_waste_disagg_matrices.py` are updated to load the expected config from the real YAML via `USAConfig.model_validate` and to patch `materialize_electricity_disagg_cornerstone_frames` directly instead of individual `derive_*` functions.

PR #417, opened 2026-05-16
Explicitly use 2025_usa_cornerstone_full_model.yaml as the config file for electricity disagg requisite matrix generation

jvendries:
cc:
Closes:

## What changed? Why?

The precondition check in `export_cornerstone_waste_disagg_matrices.py` previously maintained a hardcoded duplicate of the flags from `2025_usa_cornerstone_full_model.yaml` (`_FULL_MODEL_MATRIX_EXPORT_FLAGS`). This meant that any edit to the YAML would silently drift out of sync with the code. The hardcoded dict has been replaced with `_load_required_yaml_pins()`, which reads the canonical YAML directly at runtime so the YAML itself is the single source of truth for the precondition.

Additional changes:
- `clear_publish_caches()` is now called inside `export_cornerstone_matrices_to_csv()` before any derive calls, ensuring stale cached data does not contaminate exports.
- The `__main__` block now respects an already-set `USA_CONFIG_ENV_VAR` instead of unconditionally overwriting it.
- The mismatch error message now includes `_REQUIRED_CONFIG_FILE` so callers know exactly which YAML to load.
- `MagicMock(spec=USAConfig)` is used in tests so unknown attribute access raises `AttributeError` rather than returning a new mock, catching typos in pin names at test-construction time.

## Testing

- Precondition asserts USA_CONFIG_FILE == 2025_usa_cornerstone_full_model.yaml plus loadable waste weights, instead of comparing individual config flags against the YAML.
- Exporter-specific unit tests removed per review; one test retained in tests_electricity_disagg.py for the PR1 Y thin-wrapper used by cornerstone_Ytot_full_cs.csv export.


jvendries:
* **#417** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/417?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/417?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#395** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/395?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#411](https://github.com/cornerstone-data/bedrock/pull/411) <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/411?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

jvendries:
@bl-young I'd appreciate your review on the tests file to trim it down if you find some tests unnecesary

bedrock/extract/disaggregation/export_cornerstone_waste_disagg_matrices.py
```diff
@@ -16,33 +22,49 @@
...
+        loaded = yaml.safe_load(f) or {}
+    if not isinstance(loaded, dict):
+        raise RuntimeError(
+            f"{_REQUIRED_CONFIG_FILE} did not parse to a mapping; got "
+            f"{type(loaded).__name__}"
+        )
+    return loaded
 
 
 def assert_cornerstone_matrix_export_preconditions() -> None:
```
bl-young, line 29: especially because this file is not meant to integrate into the pipeline, but rather to be run one time to generate the necessary electricity input files, I think this feels over protective. I think including something somewhere like the following to confirm that assertion should be sufficient rather than checking individual config flags:
```
if os.environ.get(USA_CONFIG_ENV_VAR) != _REQUIRED_CONFIG_FILE:
```

bl-young (commented):
Regarding testing, I think because this is a once-run script and not a pipeline integrating functions I'm not sure really any tests are needed. There's not really much risk that something is going to change in `export_cornerstone_waste_disagg_matrices.py` unless we intend it to - at which point we'd have to make an associated update in the test file.

The `test_derive_cornerstone_ytot_full_cs_matrix_is_copy_of_underlying` could perhaps stay and instead go in a different test file.

bedrock/extract/disaggregation/export_cornerstone_waste_disagg_matrices.py
bl-young: So i guess more concretely, drop the `load_required_yaml_pins()` piece and just assert it directly. Keeping ` if get_waste_disagg_weights() is None...` seems fine.

jvendries: Implemented [here](https://github.com/cornerstone-data/bedrock/blob/Electricity_disagg_PR_1_2/bedrock/extract/disaggregation/export_cornerstone_waste_disagg_matrices.py#L24-L32)

bl-young (approved):
Approved as follow on to #395

jvendries:
### Merge activity

* **May 18, 8:28 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/417).
* **May 18, 8:29 PM UTC**: [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/417) rebased this pull request as part of a merge.
* **May 18, 8:32 PM UTC**: @jvendries merged this pull request with [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/417).

PR #420, opened 2026-05-19
Implement electricity reallocation

### LINKED ISSUE #83: Electricity re-allocation: treatment of IO values and emissions when reallocating the electricity co-production values
**[Decision](https://github.com/cornerstone-data/methods/issues/83#issuecomment-4491865282):** Leave `E` unchanged during the reallocation. During the actual electricity disaggregation step (i.e. after reallocation) use eGRID data to match emissions to power generation and use GHGI for the transmission and distribution sectors.

This issue will be closed when the reallocation [PR](https://github.com/cornerstone-data/bedrock/pull/420) is fully implemented and tested

**Problem (conceptual)**

In order to integrate physical units in the IO tables for the electricity sector, the electricity sector (`221100`) in the Make table should not co-produce any other commodities, and no other industries should co-produce the electricity commodity. Otherwise, we will not be able to compute `x` and `q` without mixing monetary and physical units, which blocks running the rest of the pipeline.

As a result, this step is meant to reallocate co-production in Make by:

1. Reallocating any `221100` commodity production by non-`221100` industries to the `221100` industry so it sits along the `221100` diagonal (i.e., movement to the `221100` row/column intersection in the Make table).
2. Reallocating any non-electricity commodity produced by the `221100` industry to that commodity's primary industry so it sits on that primary industry's diagonal (i.e., movement to the non-`221100` row/column intersections the Make table).
3. In the Use table (and VA), copy the prorportional industry movements that happen in the Make table. I.e., reallocating the supply chain inputs of reallocated industries proportionately so that the industry totals match between the Make and Use tables. This constiutes movements between the different columns of the Use table for all rows that have non-zero values in the supply chains of the reallocated industries. 

For a concrete example, see Tab 2: "Realloc Electricity" in this [spreadsheet](https://docs.google.com/spreadsheets/d/1Pi2sll_y_YiZfNjy0QNqxLEb0YcloA1M/edit?usp=drive_link&ouid=116017163929538599634&rtpof=true&sd=true).

The expected outcome of this step is the ability to run a full cornerstone pipeline from those reallocated tables, producing Aq, B, L, M, N, etc.

**Main question:** 

Should emissions totals (e.g., `E`) be reallocated along with industry totals (i.e., the `U` + `VA` movement)? 

**Option A:** Reallocate E with Use + VA (same column rules)

Idea: Treat E like any other industry-attributed satellite row: when activity (and its intermediate inputs and value added) is reassigned between industry columns, the same share of direct emissions should be reassigned.

**Pros**

_Interpretability_: E stays aligned with “who the model says produced the output” after co-production cleanup. Direct emissions intensity Eᵢ / xᵢ refers to the same industry column as U and VA, where `E` is the direct industry emissions matrix returned by [`derive_E_usa()`](https://github.com/cornerstone-data/bedrock/blob/main/bedrock/transform/allocation/derived.py#L196) and `x` is the Make table row sums, i.e., 

$$ x_i =  \sum_{j \in J} V_{ij}  $$

_Consistency with the reallocation method_: This step is explicitly about moving economic attribution between industry columns without changing commodity-level totals in ways that break supply–use logic. Moving `E` preserves the meaning “emissions are tied to the same activity column as purchases and value added.”
Downstream products that multiply `E`-like objects by IO activity (e.g. `B`-style constructions, footprint attributions by industry) stay internally coherent with the reallocated `U/VA` narrative, without requiring a separate rule for “satellite vs monetary tables.”

**Cons**

Emissions totals can be from a different year than the BEA Use table. Reallocating `E` is a modeling choice (proportional column shift), not something Flowsa observed. That should be documented and, if needed, sensitivity-tested (e.g. if `E` granularity does not match U assumptions).

Implementation must ensure row semantics of `E` (gas / flow buckets) are unchanged—only columns (industries) move mass like `Use/VA`.

**Option B:**  Leave `E` unchanged

Idea: Only IO-side `V/U/VA/Y` are reallocated; `E` remains exactly as produced by the current `derive_E_usa / Flowsa` path for the pre-realloc industry attribution.

**Pros**

_Data provenance:_ `E` is not from the same source as `U`; keeping it unchaged avoids imposing IO reallocation rules on GHG inventory structure.

_Less implementation risk_ on the emissions side in the short term.

**Cons**

_Column mismatch_:  After reallocation, some industry columns have different levels of activity that the original `U/VA` attribute to it, but `E` still attributes direct emissions to those original levels of activity. So `E` and `U/VA` no longer describe the same industry-level activity split, and direct emissions intensity might be affected. 

_Explainability_: This approach will require us to explain the difference in treatment between the IO and emissions side of the model. I.e., anything that combines industry-level `E` with industry-level activity from the reallocated `U` such as `B` becomes cumbersome to explain: the numerator (`E`) and the economic column definitions (`U, VA`) refer to different attributions.

**Other options?**

jvendries:
cc: https://github.com/cornerstone-data/methods/issues/79,
Closes: https://github.com/cornerstone-data/methods/issues/83

## What changed? Why?

Adds a new `implement_electricity_reallocation` feature flag that reallocates 221100 electricity co-production off-diagonal entries in the Make table to their respective diagonal positions, with corresponding adjustments propagated to the Use (domestic and import) and Value Added tables. Final demand (Y) is left unchanged.

The reallocation works by building an ordered transfer schedule from the Make matrix: inbound transfers (other industries → 221100 diagonal) are applied first, followed by outbound transfers (221100 row → other commodity diagonals), each sorted by magnitude. Each transfer proportionally shifts the corresponding columns in Udom, Uimp, and VA to preserve row totals. After all transfers, sparsity of the 221100 off-diagonals is asserted.

The flag requires `implement_waste_disaggregation: true` and is validated at config load time. Two new config files are added: a full-model config with electricity reallocation enabled and a test config. The `derive_cornerstone_V`, `derive_cornerstone_U_with_negatives`, and `derive_cornerstone_VA` functions now branch on `electricity_reallocation_enabled()`, delegating to a cached `_derive_cornerstone_io_after_electricity_reallocation()` bundle when the flag is active. The pre-reallocation (waste-only) derivation logic has been extracted into private helpers (`_derive_cornerstone_V_after_waste`, `_derive_cornerstone_U_after_waste`, `_derive_cornerstone_VA_after_waste`) to avoid duplication.

A CLI script (`write_221100_reallocation_matrices.py`) and supporting analysis module (`reallocation_matrices.py`) are added to export before/after Make and extended Use tables to a seven-sheet Excel workbook for inspection.

Call as follows:

`uv run python -m bedrock.analysis.electricity.write_221100_reallocation_matrices --config_name test_usa_config_waste_disagg_electricity.yaml`

or 

`uv run python -m bedrock.analysis.electricity.write_221100_reallocation_matrices --config_name 2025_usa_cornerstone_full_model_electricity_reallocation.yaml`

## Testing

- Unit tests cover transfer schedule ordering, zero-row-sum error handling, and single-transfer row-total preservation.
- Integration tests (marked `eeio_integration`) verify post-reallocation Make diagonal values match expected totals, commodity balance for 221100 is preserved, final demand for 221100 is unchanged, the emissions satellite matrix `E` is unaffected, the full pipeline produces correctly dimensioned `Aq` matrices (405×405), and the feature-off regression confirms unrelated sectors are unchanged.
- A config-parsing test confirms `implement_electricity_reallocation` is correctly read, and a validation test confirms the flag cannot be enabled without `implement_waste_disaggregation`.


jvendries:
* **#420** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/420?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/420?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jvendries:
There are several tests in `\transform\eeio\__tests__\test_electricity_reallocation.py` such as `test_e_unchanged_with_electricity_reallocation` that we might want to remove once the disaggregation is completely implemented and tested, but that right now serve as a good validation.

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -339,27 +413,13 @@ def derive_cornerstone_Vnorm_scrap_corrected(
...
-
-    weights = get_waste_disagg_weights()
-    if weights is not None:
-        Udom_cs, Uimp_cs = apply_waste_disagg_to_U(Udom_cs, Uimp_cs, weights)
-        for df in (Udom_cs, Uimp_cs):
-            df.index.name = 'sector'
-            df.columns.name = 'sector'
-
+    if electricity_disaggregation_enabled():
+        b = _derive_cornerstone_io_after_electricity_reallocation()
```
bl-young: is there a reason why you are using `b` here?

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -687,6 +724,7 @@ def derive_cornerstone_B_via_vnorm() -> pd.DataFrame:
     """B (ghg × Cornerstone commodity).
 
     Always computed in Cornerstone space: E = derive_E_usa(), then B = (E / x) @ Vnorm.
+    E is not column-shifted by electricity reallocation (PR2); Vnorm and x use reallocated V.
```
bl-young: I think this comment adds more confusion that it solves and feels out of place. Without context in this PR I think it won't be clear what this is about. I think it's fine to remove entirely.

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -189,15 +193,24 @@ def get_waste_disagg_weights() -> DisaggWeights | None:
...
 
-# ---------------------------------------------------------------------------
-# Base 2017 IO matrices — V, g, q
-# ---------------------------------------------------------------------------
+@functools.cache
+def electricity_disaggregation_enabled() -> bool:
+    cfg = get_usa_config()
+    return (
+        cfg.implement_electricity_disaggregation
+        and get_waste_disagg_weights() is not None
```
bl-young: this double boolean shouldn't need to be evaluated again if you already have the assertion in usaconfig.

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -189,15 +193,24 @@ def get_waste_disagg_weights() -> DisaggWeights | None:
...
+    )
 
 
-@functools.cache
-@pa.check_output(CornerstoneVMatrix.to_schema())
-def derive_cornerstone_V() -> pd.DataFrame:
-    """V matrix (industry × commodity) via correspondence multiplication."""
+@dataclass
+class _CornerstoneIOBundle:
```
bl-young, line 203: is this class something we intend to use moving forward? If not, i think it's superfluous just to bundle these objects together. I see it is only used in one spot at the moment.

bedrock/transform/eeio/electricity_disaggregation.py
```diff
@@ -0,0 +1,212 @@
...
+
+def check_y_unchanged(
+    Y_before: pd.DataFrame,
+    Y_after: pd.DataFrame,
+    commodity: str = ELECTRICITY_AGGREGATE,
+    atol: float = 1e-6,
+) -> pd.DataFrame:
+    """Return Y_after, warning and proportionally adjusting FD if row drift exceeds atol."""
+    if commodity not in Y_before.index or commodity not in Y_after.index:
+        return Y_after.copy()
```
bl-young: would there be a case where this function would be run with a `commodity` that does not exist in `Y`? If so should that deserve more than a silent return?

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -210,6 +223,67 @@ def derive_cornerstone_V() -> pd.DataFrame:
...
+    VA.columns.name = 'sector'
+    weights = get_waste_disagg_weights()
+    if weights is not None:
+        VA = apply_waste_disagg_to_VA(VA, weights)
+        VA.columns.name = 'sector'
+    return VA
+
+
+@functools.cache
+def _derive_cornerstone_io_after_electricity_reallocation() -> _CornerstoneIOBundle:
```
bl-young, line 259: It took me a moment, but I'm following along with your branching here. Electricity requires that all these pieces collectively have to be available, V, U, y, VA, so you can't handle each one one-by-one but rather need to derive them all first using the main pipeline before reallocating. Because this is cached, to get `V` w/ electricity reallocated all the others will have to be created anyway (`U`, `y`, `VA`) the first time through.

bedrock/transform/eeio/electricity_disaggregation.py
```diff
@@ -0,0 +1,212 @@
...
+
+@dataclass(frozen=True)
+class CoprodTransfer:
+    source: str
+    target: str
+    amount: float
+
+
+def build_coproduction_transfer_schedule(V: pd.DataFrame) -> list[CoprodTransfer]:
+    """Build ordered transfer list for 221100 row/column off-diagonals."""
```
bl-young: can you explain more what this function is doing? Is phase 1 industries and phase 2 commodities? if so can we call them that instead?

bedrock/transform/eeio/electricity_disaggregation.py
```diff
@@ -0,0 +1,212 @@
...
+            f"row_max={float(row_off.abs().max())}, "
+            f"col_max={float(col_off.abs().max())}"
+        )
+
+
+def expected_post_reallocation_diagonals(
+    V_pre: pd.DataFrame,
+    aggregate: str = ELECTRICITY_AGGREGATE,
+) -> pd.Series[float]:
+    """Expected diagonal Make values after full reallocation (static schedule)."""
```
bl-young: I'm not sure I understand this function. Where is it used?

bedrock/transform/eeio/electricity_disaggregation.py
```diff
@@ -0,0 +1,212 @@
...
+                expected[aggregate] += t
+    for d in V_pre.columns:
+        if d != aggregate:
+            t = _frame_cell_float(V_pre, aggregate, str(d))
+            if t > 0:
+                expected[d] += t
+    return expected
+
+
+def check_y_unchanged(
```
bl-young: I don't see this used anywhere. I would think we would want to actively adjust Y if needed and not do it in combination with a review of some kind.

bedrock/transform/eeio/__tests__/test_waste_disagg_pipeline_integration.py
```diff
@@ -16,9 +16,12 @@
 
 from bedrock.extract.disaggregation.disagg_weights import DisaggWeights
 from bedrock.transform.allocation.derived import derive_E_usa
-from bedrock.transform.eeio import cornerstone_expansion
+from bedrock.transform.eeio import (
+    cornerstone_expansion,
+)
 from bedrock.transform.eeio.derived_cornerstone import (
     _WASTE_NEW_CODES,
+    _derive_cornerstone_io_after_electricity_reallocation,
```
bl-young: I'm not seeing why these imports were added in this file

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
```diff
@@ -0,0 +1,297 @@
...
+        ]
+        np.testing.assert_allclose(
+            _float_ndarray(V_full.loc[unrelated, unrelated].to_numpy()),
+            _float_ndarray(V_waste.loc[unrelated, unrelated].to_numpy()),
+            rtol=1e-9,
+            atol=1e-6,
+        )
+
+    @pytest.mark.eeio_integration
+    def test_pipeline_b_dimensions(self) -> None:
```
bl-young: these last two tests feel like they can be dropped IMO

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
```diff
@@ -0,0 +1,297 @@
...
+        )
+
+
+@pytest.mark.eeio_integration
+class TestElectricityReallocationIntegration:
+    def teardown_method(self) -> None:
+        _teardown()
+
+    def test_reallocation_clears_221100_off_diagonals(self) -> None:
+        _setup_config("test_usa_config_waste_disagg_electricity.yaml")
```
bl-young: am I correct in thinking that the main pipeline already has this assertion in it and so a new separate test would just confirm the same thing?

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
```diff
@@ -0,0 +1,297 @@
...
+        for industry in touched:
+            np.testing.assert_allclose(
+                _make_diagonal(V_post, industry),
+                expected.loc[industry],
+                rtol=0,
+                atol=BALANCE_TOLERANCE,
+            )
+
+    def test_commodity_221100_totals_preserved(self) -> None:
+        _setup_config("test_usa_config_waste_disagg.yaml")
```
bl-young: same with this one is this already checked in the main pipeline?

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
```diff
@@ -0,0 +1,297 @@
...
+    _clear_all_caches()
+    reset_usa_config(should_reset_env_var=True)
+
+
+class TestElectricityConfig:
+    def teardown_method(self) -> None:
+        _teardown()
+
+    def test_electricity_flag_requires_waste(self) -> None:
+        with pytest.raises(
```
bl-young: I think this duplicates the usaconfig assertion

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
```diff
@@ -0,0 +1,297 @@
+"""Tests for 221100 electricity co-production reallocation (PR2)."""
```
bl-young: probably can drop references to PR2

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
```diff
@@ -0,0 +1,297 @@
...
+                {
+                    "implement_electricity_disaggregation": True,
+                    "implement_waste_disaggregation": False,
+                },
+                strict=True,
+            )
+
+    def test_electricity_disagg_config_parsing(self) -> None:
+        config = _load_usa_config_from_file_name(
+            "test_usa_config_waste_disagg_electricity.yaml"
```
bl-young: since this is testing against the test config this tests feels superfluous

bl-young (commented):
Provided some inline minor comments but for the most part I am tracking the approach. Is it possible to produce csv versions of the matrices after the reallocation so we can see that its working as intended? To me that's the most important validation/proof of this step which is hard to track in the code itself.

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
jvendries: Dropped references in https://github.com/cornerstone-data/bedrock/pull/420/commits/e13d28b815860b28336dabe13609a3b2e94fcf8c

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
jvendries: Removed the duplicate `TestElectricityConfig` class from `test_electricity_reallocation.py` and dropped the unused `USAConfig / _load_usa_config_from_file_name` imports in https://github.com/cornerstone-data/bedrock/pull/420/commits/c401e64a902a79f2838559d3dc1617c8e909ebee

Config coverage stays in `test_usa_config.py`:

`test_electricity_disagg_requires_waste`
`test_electricity_disagg_config_parsing`

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
jvendries: Removed in https://github.com/cornerstone-data/bedrock/commit/c401e64a902a79f2838559d3dc1617c8e909ebee

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
jvendries: Yes, you are right. 
Removed `test_reallocation_clears_221100_off_diagonals` from `TestElectricityReallocationIntegration` in https://github.com/cornerstone-data/bedrock/pull/420/commits/c401e64a902a79f2838559d3dc1617c8e909ebee

221100 off-diagonal sparsity is still enforced in production by `assert_221100_make_sparsity()` at the end of `reallocate_electricity_coproduction()`. Any integration test that successfully calls derive_cornerstone_V() with the electricity config already calls that assert.

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
jvendries: No, this is not checked in the main pipeline. 

`test_commodity_221100_totals_preserved` checks that `221100` commodity totals are preserved after reallocation with respect to the pre-reallocation totals in both the Make and Use + Y tables. 

The main pipeline checks 1) 221100 row/column off-diagonals ≈ 0 after reallocation via `assert_221100_make_sparsity(V)`; 2) that the commodity totals are unchanged after each step of the reallocation algorithm via `_assert_row_totals_unchanged`. 

However `test_commodity_221100_totals_preserved` is almost the same test as `test_make_use_commodity_balance`, so it is dropped in https://github.com/cornerstone-data/bedrock/pull/420/commits/c401e64a902a79f2838559d3dc1617c8e909ebee.

bedrock/transform/eeio/__tests__/test_electricity_reallocation.py
jvendries: Agree we can drop `test_pipeline_b_dimensions` as it only really checks `B` row count.

I think we should keep `test_diagnostics_helpers_run` as this test is the only automated test that loads `2025_usa_cornerstone_full_model_electricity_disagg.yaml` and runs the full diagnostics path (`pull_efs_for_diagnostics()` (`B`, `L`/`M`/`N`, etc.).

Edit: dropped `test_pipeline_b_dimesions` in https://github.com/cornerstone-data/bedrock/pull/420/commits/c401e64a902a79f2838559d3dc1617c8e909ebee

bedrock/transform/eeio/__tests__/test_waste_disagg_pipeline_integration.py
jvendries: Removed the electricity imports in https://github.com/cornerstone-data/bedrock/pull/420/commits/3373b3d37e451cedbea06e42d95637f95866a6c4

bedrock/transform/eeio/derived_cornerstone.py
jvendries: Agreed. Simplified to `return cfg.implement_electricity_disaggregation` in https://github.com/cornerstone-data/bedrock/pull/420/changes/d636760037aac7458e0049de0458bf9b5e1143a5.  `USAConfig` guarantees the waste prerequisite.

bedrock/transform/eeio/derived_cornerstone.py
jvendries: I think we should keep it, as it is used in various places, specifically:
- In `derive_cornerstone_V`:  `return derive_cornerstone_io_after_electricity_reallocation().`**V.copy()**
- In `derive_cornerstone_U_with_negatives()`:  `Udom=pt.DataFrame[CornerstoneUMatrix](reallocated_io`.**Udom.copy()**`)` (and same for `U.imp`)
-  `derive_cornerstone_VA`:       `return _derive_cornerstone_io_after_electricity_reallocation()`.**VA.copy()**

All three share `_derive_cornerstone_io_after_electricity_reallocation()` via `@functools.cache`, so the first call to any of them builds the full reallocated set once, which is intended.

bedrock/transform/eeio/derived_cornerstone.py
jvendries: Yes, all pieces have to be collectively available in memory so they are all created the first time through

bedrock/transform/eeio/derived_cornerstone.py
jvendries: Good catch. Renaming `b` to ` reallocated_io` in https://github.com/cornerstone-data/bedrock/pull/420/commits/6efc86790cdd57a130b6bf3c5e436f56aeb22c4d

bedrock/transform/eeio/derived_cornerstone.py
jvendries: Removed comment in https://github.com/cornerstone-data/bedrock/pull/420/commits/d636760037aac7458e0049de0458bf9b5e1143a5

bedrock/transform/eeio/electricity_disaggregation.py
jvendries: This function is creating an ordered list of transfers of the electricity re-allocations which are carried out in `reallocate_electricity_coproduction()` one at a time. 

`Phase` 1 are all the transfers where `221100` is the the _target_ industry for the transfer, ordered from largest to smallest transfer value within this subset.
`Phase` 2 are all the transfers where `221100` is the _source_ industry for the transfer, ordered from largets to smallest transfer value within this subset.

I can rename Phase 1 and 2 to `inbound_to_221100_diagonal ` and `outbound_from_221100_diagonal`, respectively.

The order matters here in two ways: 

1) The movements for all tables (Make, Use, VA) have to be done for each step before the next movement for any of these tables can be done, or else the totals will not match; 

2) As currently structured, each transfer of Phase 1 is first applied to the Make table values of non-electricity industries (rows) producing electricity commodity and moving those values _to_ the `221100` industry, along the diagonal (i.e., moving between make table rows). Then a ratio is calculated: 

`R = transfer/ 221100 Make row sum`

This ratio is used to move the same total value in the Use+VA table between industries (columns). Values are moved from all non-zero elements of the column, but the total movement is the same between industries in the Make and Use table. 

Phase 2 then moves values non-electricity commodities _from_ the `221100` industry (row) to other industries (Nat Gas distribution, other real estate) in the Make table, and matches that movement in the the Use table by moving `221100` column values matching that total to other columns. 

If Phase 1 happens first, the denominator in the ratio `R` above is larger, resulting in smaller transfers out of the `221100` industry column in absolute value. I think this is what we want; the order can be reversed if we needed, but the industry totals we be different.

bedrock/transform/eeio/electricity_disaggregation.py
jvendries: This function is a test helper for `test_post_reallocation_diagonal_values` in `test_electricity_reallocation.py`, but it is not in the production pipeline.

I moved it to `test_electricity_reallocation.py`

bedrock/transform/eeio/electricity_disaggregation.py
jvendries: You're right, this isn't called anywhere. We can remove

bedrock/transform/eeio/electricity_disaggregation.py
jvendries: No longer relevant since we will remove the function

jvendries:
> Provided some inline minor comments but for the most part I am tracking the approach. Is it possible to produce csv versions of the matrices after the reallocation so we can see that its working as intended? To me that's the most important validation/proof of this step which is hard to track in the code itself.

I added the functionality to print the reallocated Make and Use tables after reallocation under `\bedrock\utils\validation\analysis\electricity_disagg_intermediate_outputs\` and I am attaching the output workbook here.

[V_and_U_after_221100_reallocation.xlsx](https://github.com/user-attachments/files/28371494/V_and_U_after_221100_reallocation.xlsx)

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -210,6 +220,76 @@ def derive_cornerstone_V() -> pd.DataFrame:
...
+    Udom, Uimp = _derive_cornerstone_U_after_waste()
+    VA = _derive_cornerstone_VA_after_waste()
+    V, Udom, Uimp, VA = reallocate_electricity_coproduction(V, Udom, Uimp, VA)
+    bundle = _CornerstoneIOBundle(V=V, Udom=Udom, Uimp=Uimp, VA=VA)
+
+    from bedrock.utils.validation.analysis.electricity_disagg_intermediate_outputs import (  # noqa: PLC0415
+        maybe_write_electricity_disagg_intermediate_outputs,
+    )
+
+    maybe_write_electricity_disagg_intermediate_outputs(
```
bl-young: it seems strange to me to pull this function in to the core pipeline. couldn't we instead call `_derive_cornerstone_io_after_electricity_reallocation()` and then call the writer function in a separate script?

bedrock/utils/validation/analysis/electricity_disagg_intermediate_outputs/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Write post-reallocation Make / extended Use artifacts for electricity disagg."""
```
bl-young: most of the items in `validation/analysis` are post-pipeline validation runs for diagnostics. I would propose we move these functions related to assessing this stage of electricity disaggregation to `bedrock/analysis/electricity/`

bedrock/utils/validation/analysis/electricity_disagg_intermediate_outputs/writer.py
```diff
@@ -0,0 +1,200 @@
...
+    intermediate = udom_nonneg + uimp_nonneg
+    extended = _assemble_extended_U(
+        intermediate=intermediate,
+        fd=y_fd,
+        va=va.copy(),
+    )
+    return extended, intermediate
+
+
+def maybe_write_electricity_disagg_intermediate_outputs(
```
bl-young: I would instead propose this as a separate script or a command-line script to call this function, and then you could drop the `maybe_` at the front

bl-young (approved):
Looks good - see my suggestion regarding the approach to writing the output (thanks for that!). I'll mark this as approved pending further discussion of how to approach that.

bedrock/transform/eeio/electricity_disaggregation.py
bl-young: makes sense! This might be good to include as a comment (or some of it at least) around this function!

bedrock/transform/eeio/derived_cornerstone.py
jvendries: Removed from core pipeline

bedrock/utils/validation/analysis/electricity_disagg_intermediate_outputs/__init__.py
jvendries: Moved to proposed location under `bedrock/analysis/electricity/`

bedrock/utils/validation/analysis/electricity_disagg_intermediate_outputs/writer.py
jvendries: added separate script, called via 

`uv run python -m bedrock.analysis.electricity.write_221100_reallocation_matrices --config_name test_usa_config_waste_disagg_electricity.yaml`

or for full model config run:

`uv run python -m bedrock.analysis.electricity.write_221100_reallocation_matrices --config_name 2025_usa_cornerstone_full_model_electricity_reallocation.yaml`

bedrock/transform/eeio/electricity_disaggregation.py
jvendries: Added a comment under the function, [here](https://github.com/cornerstone-data/bedrock/pull/420/changes#diff-e28b85cc1e98367db80471362e42d58a2cf6ced11cf78b304c89f225684c1592R32-R46)

bl-young:
thanks looks good!

PR #437, opened 2026-05-26
Add functions to load 2017 margins before and after redef

### LINKED ISSUE #428: Load Margins
load from GCS

WesIngwersen:
Closes: #428 

## What changed? Why?
matrix_mappings.py — Added "Margins" to `USA_2017_DETAIL_IO_MATRIX_NAMES` and to `USA_2017_DETAIL_IO_BEFORE_REDEF_MATRIX_MAPPING` so the existing `_load_2017_detail_make_use_usa `helper accepts it without a type error. 

io_2017.py — 

Added load_2017_margins_usa() following the same pattern as the other loaders which either calls 
`load_2017_margins_after_redef_usa` or `load_2017_margins_before_redef_usa` depending on model config
 - @functools.cache for memoization just for latter two functions
- Use common `_load_2017_margins_from_file(filename)` (same GCS bucket, same Excel loading logic) to load the appropriate file
- Scales from million USD → USD via MILLION_CURRENCY_TO_CURRENCY
- skip the 5 title/header rows, assign the column names manually
- set_index(["Industry Code", "Commodity Code"]) — creates a MultiIndex so each (industry, commodity) pair is a unique row
- dtype={"Industry Code": str, "Commodity Code": str} — both code columns preserved as strings
- warnings are handled by _load_margins_excel, which suppresses the openpyxl warning with a targeted filter scoped to that with block — it won't silence warnings anywhere else in the codebase. The loader reference in load_from_gcs is also cleaner now that it's a named function rather than a lambda.
- uses a multiindex that combines industry code and commodity code columns to make each row unique in df
- valid_industry — union of USA_2017_INDUSTRY_CODES and USA_2017_FINAL_DEMAND_CODES, covering all valid industry-side codes in the BEA taxonomy
- valid_commodity — union of USA_2017_COMMODITY_CODES and USA_2017_VALUE_ADDED_CODES, covering all valid commodity-side codes
- Both sets are applied as a combined & mask across their respective MultiIndex levels — the same intent as .loc[VALID_ROWS, VALID_COLS] in the single-index functions, just adapted for the MultiIndex structure here
 
 ## Testing
Manual calling and inspection of returned df
```
from bedrock.extract.iot.io_2017 import load_2017_margins_after_redef_usa
m = load_2017_margins_after_redef_usa()
```
and to test before redef loading
```
from bedrock.extract.iot.io_2017 import load_2017_margins_usa
from bedrock.utils.config.usa_config import set_global_usa_config

set_global_usa_config("useeio_phoebe_23")
m = load_2017_margins_usa()
```

### DIFF
--- a/bedrock/extract/iot/io_2017.py
+++ b/bedrock/extract/iot/io_2017.py
@@ -276,14 +276,14 @@ _MARGINS_COLUMNS = [
     "Commodity Code",
     "Commodity Description",
     "Producers' Value",
-    "Transportation Costs",
+    "Transportation",
     "Wholesale",
     "Retail",
     "Purchasers' Value",
 ]
 _MARGINS_VALUE_COLUMNS = [
     "Producers' Value",
-    "Transportation Costs",
+    "Transportation",
     "Wholesale",
     "Retail",
     "Purchasers' Value",
--- a/bedrock/extract/iot/io_2017.py
+++ b/bedrock/extract/iot/io_2017.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import functools
+import warnings
 
 import pandas as pd
 from typing_extensions import deprecated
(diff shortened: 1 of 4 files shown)


WesIngwersen:
superseded by #440

PR #447, opened 2026-06-01
Additional USEEIO baseline updates

bl-young:
cc:
Closes:
### 1. Before-redefinition waste disaggregation weights (USEEIO parity)

USEEIO-based configs pair **before-redefinition** 2017 IO with the waste disaggregation spec shipped in **useeior v1.8.0**. Bedrock’s bundled `waste_disagg_inputs/WasteDisaggregationDetail2017_*.csv` files differ from that spec as they reflect after-redefinition.

**Behavior**

- `effective_waste_disagg_config()` in `bedrock/extract/disaggregation/waste_weight_config.py` selects weight files from `USAConfig`:
  1. **`before` IO** → USEEIOR v1.8.0 Make/Use CSVs (`useeior_v1_8_waste_disagg_config()`).
  2. **Explicit `eeio_waste_disaggregation` in YAML** → config block (tests and overrides).
  3. **`after` IO, no YAML block** → bundled Cornerstone CSVs under `waste_disagg_inputs/`.
- `bedrock/extract/disaggregation/useeior_waste_weights.py` downloads useeior v1.8.0 files on first use into `bedrock/extract/disaggregation/waste_disagg_inputs/useeior_v1.8.0/` (cached on disk).

### 2. CH4 GWP for `GHG_national_2023_m2` (USEEIO workbook alignment)

`load_E_from_flowsa()` in `bedrock/transform/allocation/derived.py` builds sector `E` from Flow-By-Sector GHG data. CEDA AR6 mapping splits `CH4_fossil` and `CH4_non_fossil` with distinct GWPs. The USEEIO workbook applies a **single CH4 factor (27.9)** for both species when characterizing the FBS used in that baseline.

### Other changes on branch (secondary)

- `useeio_phoebe_23_restore_cornerstone_A.yaml`: `model_base_year: 2023` for A-matrix targeting in that sensitivity (orthogonal to waste/CH4).

## Testing

New tests pass.


bedrock/extract/disaggregation/waste_weight_config.py
```diff
@@ -0,0 +1,43 @@
...
+    """Bedrock-bundled waste disagg weights (after-redefinition default)."""
+    return EEIOWasteDisaggConfig(
+        use_weights_file=f"{WASTE_INPUTS_REL}/{WASTE_DISAGG_USE_FILENAME}",
+        make_weights_file=f"{WASTE_INPUTS_REL}/{WASTE_DISAGG_MAKE_FILENAME}",
+        year=CORNERSTONE_WASTE_YEAR,
+        source_name=CORNERSTONE_WASTE_SOURCE_NAME,
+    )
+
+
+def effective_waste_disagg_config(cfg: USAConfig) -> EEIOWasteDisaggConfig:
```
jvendries, line 27: This function always returns USEEIOR v1.8.0 when `iot_before_or_after_redefinition == "before"`, even if `eeio_waste_disaggregation` is set in a YAML config. That’s consistent with the PR description, but it means before-IO configs overrides explicit YAML config calls. If intended, it's worth a comment to emphaize that a YAML config can't override the before-IO USEEIOR path.

jvendries (approved):

bedrock/extract/disaggregation/waste_weight_config.py
bl-young: thanks. added an extra warning for those configs in c5cce69

PR #450, opened 2026-06-01
Refactor estimate_suppressed fxn

catherinebirney:
cc:
Closes:

## What changed? Why?

1. Vectorized estimate_suppressed_sectors_equal_allocation() and estimate_suppressed_qcew() to speed up these functions
2. At the begining of the Cornerstone GHG FBS yaml, cache Cropland NAICS so only runs estimate suppressed data once, rather than 3 times

These changes reduce the time to locally run the 2023 GHG FBS from 43 minutes to 20 and the Employment FBS from 15 minutes to 7.

## Testing

1. Checked that FBS results are the same as before these changes for GHG Cornersone and Employment FBS.


bl-young:
I am running the integration tests on this branch to confirm no change in the FBS for 2023. Will update once its ready.

bl-young (commented):
What is your take on the feasibility of consolidating duplicate code across BLS_QCEW.py and flowbyclean.py? as it relates to filling the data suppression. Could we consolidate to a single function so that it is done consistently across both?

bl-young:
Confirming it shaved about 10 minutes off the integration tests (~18m down from ~28m) with no change in outcome 🥳 
image

catherinebirney:
> What is your take on the feasibility of consolidating duplicate code across BLS_QCEW.py and flowbyclean.py? as it relates to filling the data suppression. Could we consolidate to a single function so that it is done consistently across both?

The QCEW data is published differently than the other data we estimate (which for now is only COA NAICS). QCEW has non-naics codes that we need to retain to correctly do the parent-child mapping. And we do not add any additional child NAICS - we replace 0 values. In our general suppression function we pull in the naics crosswalk to determine parent-child relationships - and add new child NAICS to the dataset.

It would be doable to combine them - or have the BLS QCEW function call on the general function, but it would take a bit of time.

bl-young (approved):
Confirmed all GHG FBS adjusted in the same way and that no changes manifest in the final FBS.

PR #459, opened 2026-06-05
Add new q deriviation and compare q's against references

### LINKED ISSUE #457: Refine q estimate
Re-evaluate approaches to estimate cornerstone q using a redefined industry output.

WesIngwersen:
cc: @MoLi7 
Closes: #457 

## Summary

Adds a new q-estimation pathway — scaling the 2017 Cornerstone Make matrix to match BEA annual gross industry output — and an analysis script that compares it against the two existing pathways at both Cornerstone detail and BEA summary level.

***NOTE*** This PR does not implement the new q-estimation pathway into any model builds - it just creates the pathway and compares it with existing pathways.

### New functions in `derived_cornerstone.py`

- **`scale_cornerstone_V_with_authoritative_x()`**: inflates the base-2017 V matrix to `model_base_year` (optionally applying commodity PI inflation), then row-scales each industry so that V's row sums exactly match `derive_cornerstone_x_after_redefinition(model_base_year)` — the BEA annual gross industry output time series. The result is a Make matrix consistent with authoritative BEA output totals.
- **`derive_q_from_scaled_cornerstone_V_from_authoritative_x()`**: thin wrapper that calls the above and returns commodity q = `V.sum(axis=0)`.
- **`derive_cornerstone_x_after_redefinition`** gains an optional `year` parameter (defaults to `usa_ghg_data_year` when omitted, preserving existing behavior) so callers can request a specific year without re-reading config.

### New analysis script: `bedrock/analysis/q/compare_alt_q.py`

Compares three q-estimation pathways at the 405-sector Cornerstone detail level and aggregated to BEA summary level:

1. **V_scaled** — `derive_q_from_scaled_cornerstone_V_from_authoritative_x()` (this branch)
2. **Aq_scaled** — replicates `scale_a_matrix_with_summary_tables=True` + `adjust_summary_A_and_q_dollar_year=True` by calling `adjust_summary_q_dollar_year()` and `inflate_cornerstone_q_or_y_with_commodity_pi()` directly, independent of the active config
3. **inflated** — `inflate_cornerstone_q_or_y_with_commodity_pi()` applied to the base scaled q without the dollar-year rebase step

Reference benchmarks: `derive_cornerstone_x_after_redefinition(model_year)` (authoritative BEA x) and `derive_summary_q_usa(model_year)` (summary Make table q).

Outputs three matplotlib plots to `bedrock/analysis/q/output/plots/`:
- `detail_scatter_{year}.png` — 1×4 log-log scatter at Cornerstone detail level; panels 1–3 show each pathway vs authoritative x; panel 4 is a 2017 reference showing `compute_q(V)` vs `compute_x(V)` from the unmodified Make matrix
- `summary_levels_{year}.png` — ranked line plot of all five series at BEA summary level
- `summary_pct_deviations_{year}.png` — grouped bar chart of % deviation from each reference benchmark

### Tests

`test_scale_V_from_authoritative_x.py` verifies that `scale_cornerstone_V_with_authoritative_x()` row sums match the authoritative x vector to within 1e-6 relative tolerance.

## Analysis results
Aggregate totals at summary level ($M):
  x_authoritative      48,424,233,164,693
  q_make_table         48,366,332,000,000
  V_scaled             48,424,233,164,693
  Aq_scaled            48,307,281,448,622
  inflated             42,699,140,375,411

Overall pct deviation from Make Table q:
  V_scaled           +0.120%
  Aq_scaled          -0.122%
  inflated           -11.717%

Overall pct deviation from x_authoritative:
  V_scaled           -0.000%
  Aq_scaled          -0.242%
  inflated           -11.823%

detail_scatter_2023

summary_levels_2023

summary_pct_deviations_2023

### DIFF
--- a/bedrock/analysis/q/compare_alt_q.py
+++ b/bedrock/analysis/q/compare_alt_q.py
@@ -33,7 +33,7 @@ to the aggregated x_authoritative.
 
 Usage
 -----
-    python -m bedrock.analysis.compare_alt_q
+    python -m bedrock.analysis.q.compare_alt_q
 """
 
 from __future__ import annotations
--- a/bedrock/analysis/q/compare_alt_q.py
+++ b/bedrock/analysis/q/compare_alt_q.py
@@ -1,0 +1,525 @@
+"""Compare three q-estimation pathways at Cornerstone detail and BEA summary level.
+
+Pathways
+--------
+1. q_V_scaled
+   derive_q_from_scaled_cornerstone_V_from_authoritative_x()
+   V is inflated to model_base_year then rescaled so each industry row sum equals
+   the authoritative BEA gross-output x. q = V.sum(axis=0).
+
+2. q_Aq_scaled
+   scale_cornerstone_q (with dollar-year adjustment) + inflate_with_commodity_pi.
+   Replicates the derive_cornerstone_Aq_scaled() branch that is active when
+   scale_a_matrix_with_summary_tables=True AND adjust_summary_A_and_q_dollar_year=True.
+   The dollar-year flag is hardwired here so the comparison is independent of the
+   live config.
+
+3. q_inflated
+   inflate_cornerstone_q_or_y_with_commodity_pi applied directly to the base 2017 q
+   (no structural summary-table scaling). This is the commodity-price-index-only path.
+
+Reference vector
+----------------
+x_authoritative
+   derive_cornerstone_x_after_redefinition(year=model_base_year): BEA annual gross
+   industry output at model_base_year, after redefinitions, expanded to the 405-sector
+   Cornerstone schema. Already in model_base_year current dollars — no additional
+   inflation applied. Industry codes and commodity codes share the same 405-sector
+   namespace, so each q pathway is compared directly to this x.
+
+At the summary level each pathway is summed via the Cornerstone→BEA-summary mapping
+and compared both to derive_summary_q_usa(model_base_year) (the "Make Table" q) and
+to the aggregated x_authoritative.
+
+Usage
+-----
+    python -m bedrock.analysis.compare_alt_q
+"""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+
+from bedrock.transform.eeio.derived_2017 import derive_summary_q_usa
+from bedrock.transform.eeio.derived_cornerstone import (
+    derive_cornerstone_Aq,
+    derive_cornerstone_V,
+    derive_cornerstone_x,
+    derive_cornerstone_x_after_redefinition,
+    derive_q_from_scaled_cornerstone_V_from_authoritative_x,
+)
+from bedrock.utils.config.usa_config import get_usa_config
+from bedrock.utils.economic.inflation_helpers_cornerstone import (
+    adjust_summary_q_dollar_year,
+    inflate_cornerstone_q_or_y_with_commodity_pi,
+)
+from bedrock.utils.math.formulas import compute_q
+from bedrock.utils.taxonomy.bea_v2017_to_ceda_v7_helpers import (
+    load_bea_v2017_summary_to_cornerstone,
+)
+
+logger = logging.getLogger(__name__)
+
+OUTPUT_DIR = Path(__file__).parent / "output"
+PLOTS_DIR = OUTPUT_DIR / "plots"
+
+_PATHWAY_STYLE: list[tuple[str, str, str]] = [
+    ("q_V_scaled", "V_scaled", "#1f77b4"),
+    ("q_Aq_scaled", "Aq_scaled", "#ff7f0e"),
+    ("q_inflated", "inflated", "#2ca02c"),
+]
+
+
+# ---------------------------------------------------------------------------
+# Pathway builders
+# ---------------------------------------------------------------------------
+
+
+def _build_q_V_scaled() -> pd.Series:
+    """Pathway 1: q from V rescaled to authoritative gross-output x."""
+    return derive_q_from_scaled_cornerstone_V_from_authoritative_x()
+
+
+def _build_q_Aq_scaled(
+    base_q: pd.Series,
+    detail_year: int,
+    model_year: int,
+) -> pd.Series:
+    """Pathway 2: scale (dollar-year-adjusted) + inflate with commodity PI.
+
+    Replicates derive_cornerstone_Aq_scaled() with
+    scale_a_matrix_with_summary_tables=True and
+    adjust_summary_A_and_q_dollar_year=True, regardless of the active config.
+    """
+    # --- scale step (dollar-year-adjusted) ---
+    q_summary_target = derive_summary_q_usa(model_year)
+    q_summary_target_adj = adjust_summary_q_dollar_year(
+        q_summary=q_summary_target,
+        from_year=model_year,
+        to_year=detail_year,
+    )
+    q_summary_base = derive_summary_q_usa(detail_year)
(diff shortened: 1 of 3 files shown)


WesIngwersen:
I'm fixing the type error(s)

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -326,30 +326,32 @@ def _distribute_waste_parent_x_using_v_row_shares(
...
     parent total consistent with row sums of disaggregated ``V`` (same nominal
-    level as the GHG-year BEA gross output, split from 2017 Make structure).
+    level as the BEA gross output for *year*, split from 2017 Make structure).
 
     This is the industry ``x`` in ``derive_cornerstone_B_via_vnorm`` when
     ``use_E_data_year_for_x_in_B`` is True; otherwise that path uses
     ``derive_cornerstone_x()``.
     """
     cfg = get_usa_config()
+    effective_year = cfg.usa_ghg_data_year if year == 0 else year
```
bl-young: I'm curious about the choice of `usa_ghg_data_year` instead of `model_year`

bedrock/analysis/q/compare_alt_q.py
```diff
@@ -0,0 +1,524 @@
...
+   inflation applied. Industry codes and commodity codes share the same 405-sector
+   namespace, so each q pathway is compared directly to this x.
+
+At the summary level each pathway is summed via the Cornerstone→BEA-summary mapping
+and compared both to derive_summary_q_usa(model_base_year) (the "Make Table" q) and
+to the aggregated x_authoritative.
+
+Usage
+-----
+    python -m bedrock.analysis.compare_alt_q
```
bl-young:
```suggestion
    python -m bedrock.analysis.q.compare_alt_q
```

bedrock/transform/eeio/derived_cornerstone.py
WesIngwersen: That is an old choice to make it work for the useeio model... not related to this PR except a param was added to be able to override it.  It does need to be relooked at maybe in an issue outside this PR.

WesIngwersen:
@bl-young I added the text of the analysis run which includes additional summaries but note one aggregation error occuring in
`WARNING __main__: aggregate_q_to_summary: 1 sector(s) not in cs_to_summary and will be dropped (total $M 8078560285): ['331314']`

WesIngwersen:
> @bl-young I added the text of the analysis run which includes additional summaries but note one aggregation error occuring in `WARNING __main__: aggregate_q_to_summary: 1 sector(s) not in cs_to_summary and will be dropped (total $M 8078560285): ['331314']`

I'm working on this...

WesIngwersen:
> > @bl-young I added the text of the analysis run which includes additional summaries but note one aggregation error occuring in `WARNING __main__: aggregate_q_to_summary: 1 sector(s) not in cs_to_summary and will be dropped (total $M 8078560285): ['331314']`
> 
> I'm working on this...

331314 industry output is getting lost in the industry output mapping to commodity output for the comparison and the totaling.

jvendries:
> Bug fix: derived.py
load_E_from_flowsa referenced methodname (a local string extracted from a data column) as the condition for a GWP override instead of the config flag usa.use_ghg_national_2023_m2. This caused a ValueError when methodname was unassigned for certain data slices.

I think this was fixed in https://github.com/cornerstone-data/bedrock/pull/454

WesIngwersen:
> Reference

Claude did this. I guess the fix is identical character-by-character so no merge conflict is showing up, so easier just to leave it it in code again but I removed it from the description of the changes in this PR.

WesIngwersen:
> > > @bl-young I added the text of the analysis run which includes additional summaries but note one aggregation error occuring in `WARNING __main__: aggregate_q_to_summary: 1 sector(s) not in cs_to_summary and will be dropped (total $M 8078560285): ['331314']`
> > 
> > 
> > I'm working on this...
> 
> 331314 industry output is getting lost in the industry output mapping to commodity output for the comparison and the totaling.

Fixed to assign it temporarily to 331313 only in the scope of this script. Updated results. Now scaled V approach q sum = authoritative x sum.

jvendries:
@WesIngwersen thanks for doing this analysis and sharing the results, it makes the differences between approaches much easier to see.

One piece still missing for a fully informed comparison would be EF diagnostics (`D`/`N`) for the `V_scaled` and `Aq_scaled` pathways (since the latter is not quite the same as the approach for EFs we've already seen if I'm not mistaken). While aggregate `q` totals for `V_scaled` and `Aq_scaled` are close (~0.1–0.2%), I don't have a good sense of how detail-sector EFs might differ between them. Since `V_scaled` isn't yet part of a model build I imagine this would have to be a follow-on analysis.

Beyond EF impacts, each approach seems to have a distinct main benefit:

- **V_scaled:** Would allow for greater internal consistency. `x`, `q`, and `A` could be derived from the same scaled Make matrix, rather than `B` using authoritative `x` while `q`/`A` follow a separate scaling path.
- **Aq_scaled:** Continuity with the summary-table scaling framework we are currently using (even the calculation has been changed a bit via `adjust_summary_q_dollar_year`)
- **inflation:** Simplest to explain and communicate though with the largest aggregate error by a large margin if used on its own.

If EF diagnostics show relatively little difference between `V_scaled` and `Aq_scaled` at detail sector level, prioritizing among these benefits might be a reasonable way to choose an approach.

WesIngwersen:
> @WesIngwersen thanks for doing this analysis and sharing the results, it makes the differences between approaches much easier to see.
> 
> One piece still missing for a fully informed comparison would be EF diagnostics (`D`/`N`) for the `V_scaled` and `Aq_scaled` pathways (since the latter is not quite the same as the approach for EFs we've already seen if I'm not mistaken). While aggregate `q` totals for `V_scaled` and `Aq_scaled` are close (~0.1–0.2%), I don't have a good sense of how detail-sector EFs might differ between them. Since `V_scaled` isn't yet part of a model build I imagine this would have to be a follow-on analysis.
> 
> Beyond EF impacts, each approach seems to have a distinct main benefit:
> 
> * **V_scaled:** Would allow for greater internal consistency. `x`, `q`, and `A` could be derived from the same scaled Make matrix, rather than `B` using authoritative `x` while `q`/`A` follow a separate scaling path.
> * **Aq_scaled:** Continuity with the summary-table scaling framework we are currently using (even the calculation has been changed a bit via `adjust_summary_q_dollar_year`)
> * **inflation:** Simplest to explain and communicate though with the largest aggregate error by a large margin if used on its own.
> 
> If EF diagnostics show relatively little difference between `V_scaled` and `Aq_scaled` at detail sector level, prioritizing among these benefits might be a reasonable way to choose an approach.

@jvendries I will do a full N comparison in a future PR.

bedrock/analysis/q/compare_alt_q.py
```diff
@@ -0,0 +1,529 @@
...
+    summary_to_cornerstone = load_bea_v2017_summary_to_cornerstone()
+    mapping = {
+        code: str(summary)
+        for summary, codes in summary_to_cornerstone.items()
+        for code in codes
+    }
+    # BEA industry 331314 (secondary aluminum smelting) is not a commodity code;
+    # after redefinitions its output is attributed to commodity 331313.
+    if "331313" in mapping:
+        mapping["331314"] = mapping["331313"]
```
jvendries, line 153: For analysis only this mapping is OK, but is this adequate if we move this to production?

jvendries (approved):
PR looks good; I reproduced the analysis locally (`python -m bedrock.analysis.q.compare_alt_q`) and got the same aggregate totals reported in the PR description.
Agreed that full EF diagnostics (`D`/`N`) for `V_scaled` vs `Aq_scaled` belong in a follow-up PR.

PR #462, opened 2026-06-09
Implement 221100 split

jvendries:
cc:
Closes: https://github.com/cornerstone-data/methods/discussions/85

## What changed? Why?

Implements monetary disaggregation of the aggregate `221100` electricity sector into three sub-sectors — `221110` (generation), `221121` (transmission), and `221122` (distribution) — using gross-output shares derived from BEA UGO305-A data. This is the third major step in the electricity sector disaggregation pipeline, following the existing co-production reallocation work.

The disaggregation runs in four sequential steps:
1. Split the `221100` Make diagonal into a 3×3 diagonal block using GO weights
2. Split the `221100` Use self-use intersection diagonally
3. Split the `221100` industry column across sub-industries with fuel-commodity routing (generation-only fuels like coal, natural gas, petroleum go entirely to `221110`) and VA balancing to preserve row totals
4. Split the `221100` commodity row across all consuming industries using equal-price GO weights

A new `implement_electricity_disaggregation` flag is added to `USAConfig` with a validator enforcing that both `implement_waste_disaggregation` and `implement_electricity_reallocation` must be enabled as prerequisites. When active, the schema expands from 405 to 407 sectors, and all Cornerstone matrix validations (`V`, `U`, `A`, `B`, `X`, `Q`) are routed through a new flag-aware `validate_cornerstone()` helper that replaces the previous static `@pa.check_output` decorators, allowing the same validation logic to work against either the 405- or 407-sector schema depending on config.

Emissions attribution (`derive_E_usa`) loads the eGRID-based FBS (`GHG_national_Cornerstone_2023_egrid`) when electricity disaggregation is enabled, with a fallback to the cached national FBS. SF6 emissions are routed to `221121` (transmission); all other gases go to `221110` (generation). Inflation helpers, the BEA summary-to-Cornerstone correspondence, and gross-output `x` distribution are all updated to propagate the parent `221100` values to the three sub-sectors.

Two new config files are added: a test config (`test_usa_config_waste_disagg_electricity_disaggregation.yaml`) and a full production config (`2025_usa_cornerstone_full_model_electricity_disaggregation.yaml`).

## Testing

New integration tests in `test_electricity_disaggregation.py` cover:
- GO weights summing to 1.0 across the three sub-sectors
- A worked numerical example verifying column balancing, VA row preservation, and use-row totals after step 3
- Full pipeline integration tests (marked `eeio_integration`) confirming the 407-sector schema shape, absence of `221100` from all output matrices, Aq column sums ≤ 1, and non-zero emissions attribution to the disaggregated electricity columns


jvendries:
* **#508** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/508?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#492** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/492?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#490** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/490?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#481** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/481?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#463** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/463?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#462** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/462?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/462?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bedrock/utils/economic/inflation_helpers_cornerstone.py
```diff
@@ -103,13 +108,20 @@ def get_cornerstone_industry_price_ratio(
 
     # Anything still NaN (truly no parent, e.g. S00402) gets neutral 1.0
     ratio = ratio.fillna(1.0)
+    if cfg.implement_electricity_disaggregation:
```
bl-young, line 131: I'm curious how this is handled for waste right now. Because waste sectors face the same issue right? I would think that the same approach would suffice here potentially? Does the map right above with `_cornerstone_to_ceda_v7_parent()` attempt to handle that?

bedrock/utils/economic/inflation_helpers_cornerstone.py
```diff
@@ -103,13 +108,20 @@ def get_cornerstone_industry_price_ratio(
...
+            ratio.loc[code] = parent_ratio
+        ratio = ratio.reindex(CORNERSTONE_COMMODITIES_ELEC, fill_value=1.0)
     return ratio
 
 
 def inflate_cornerstone_A_matrix_with_industry_pi(
     A: pd.DataFrame, original_year: int, target_year: int
 ) -> pd.DataFrame:
     price_ratio = get_cornerstone_industry_price_ratio(original_year, target_year)
+    price_ratio = price_ratio.reindex(A.index, fill_value=1.0)
```
bl-young: this feels like it should already be done within `get_cornerstone_industry_price_ratio` so I'm not sure it needs to be done again

bedrock/utils/config/usa_config.py
```diff
@@ -79,6 +79,7 @@ class USAConfig(BaseModel):
     implement_waste_disaggregation: bool = False  # DRI: jorge.vendries
     eeio_waste_disaggregation: ta.Optional[EEIOWasteDisaggConfig] = None
     implement_electricity_reallocation: bool = False  # DRI: jorge.vendries
+    implement_electricity_disaggregation: bool = False  # DRI: jorge.vendries
```
bl-young, line 93: This comment is relevant in a few places. Are we planning to evaluate the reallocation with and without disaggregation? I'm not sure what purpose that serves really. I would propose we just have a single flag and not over burden ourselves. (to be more clear, my suggestion would be to use `implement_electricity_disaggregation` and drop `implement_electricity_reallocation`

bedrock/utils/config/__tests__/test_usa_config.py
```diff
@@ -219,6 +219,30 @@ def test_electricity_disagg_config_parsing() -> None:
     assert config.implement_electricity_reallocation is True
 
 
+def test_electricity_disaggregation_config_parsing() -> None:
```
bl-young: I think this is duplicative of the validation in usa_config.py (though per my comment about the new flag this test might not be necessary at all)

bedrock/transform/eeio/electricity_disaggregation.py
```diff
@@ -1,16 +1,66 @@
...
+    ELECTRICITY_DISAGG_SECTORS,
+)
+from bedrock.utils.taxonomy.cornerstone.value_added import VALUE_ADDEDS
+
+logger = logging.getLogger(__name__)
+
 ELECTRICITY_AGGREGATE = "221100"
 BALANCE_TOLERANCE = 1e6
+DISAGG_BALANCE_ATOL = 1.0
+IO_ACCOUNT_YEAR = 2017
```
bl-young: this should come from the model parameter inside the model config

bedrock/transform/eeio/cornerstone_disagg_pipeline.py
```diff
@@ -163,19 +176,25 @@ def derive_disagg_io_bundle() -> CornerstoneDisaggIOBundle:
     VA = derive_cornerstone_VA_after_waste()
     if electricity_reallocation_enabled():
         V, Udom, Uimp, VA = reallocate_electricity_coproduction(V, Udom, Uimp, VA)
+    if electricity_disaggregation_enabled():
```
bl-young, line 179: I have this comment elsewhere as well but i think this double gating is unnecessary

bedrock/transform/allocation/derived.py
```diff
@@ -430,12 +445,22 @@ def load_E_from_flowsa() -> pd.DataFrame:
             'update_*_ghg_method flag.'
         )
     if usa.new_ghg_method:
-        # Bypass flowsa regen for non-2023 years: the EPA loader behind
-        # `getFlowBySector` is hard-capped at {2022, 2023}, but the
-        # already-built FBS parquets exist on GCS at
-        # `transform/output_data/` for 2019–2023. Load the cached parquet
-        # directly so the year-Y diagnostics get year-Y GHG data.
-        fbs = _load_cornerstone_ghg_fbs_from_gcs(usa.usa_ghg_data_year)
+        if usa.implement_electricity_disaggregation:
```
bl-young: this implementation doesn't seem quite right. I think we should just be able to target the new method in `_select_flowsa_ghg_method` like other approaches. I think this particular if/else branch was intended for off years

bl-young (commented):
Sharing some initial comments but given the complexity I will want to review more closely.

bedrock/utils/config/usa_config.py
jvendries: Yes I agree the ultimate goal is to drop `reallocation` flag, but I do think it is worthwhile to evaluate reallocation in isolation. In other words, my intent was to keep them separate for now so we can better evaluate any source of variabity in disaggregated EFs. I.e., if the EFs change a lot, is that caused by the disaggregation itself, or the reallocation? 

Does this make sense, or do you think this separation is unnecesary even for diagnostics?

bedrock/utils/economic/inflation_helpers_cornerstone.py
jvendries: I'm not sure I fully understand the question. 

If you are asking whether we can use the same mapping as for waste, I don't think so. The waste disagg is built into the cornerstone schema, whereas the electricity disagg is not. So we need additional mapping to expand the schema. 

Does this answer the question?

bedrock/utils/economic/inflation_helpers_cornerstone.py
jvendries: Perhaps but this is the way it is structured in main. Should this be done in its own PR rather than here?

bedrock/transform/allocation/derived.py
```diff
@@ -448,12 +465,27 @@ def load_E_from_flowsa() -> pd.DataFrame:
...
     if usa.new_ghg_method or usa.v0_3_umd_2024_ghgia:
-        # Bypass flowsa regen: the EPA loader behind `getFlowBySector` is
-        # hard-capped at {2022, 2023}, so other years (incl. the 2024 UMD
-        # FBS) fail there. Load the pre-built FBS parquet from GCS at
-        # `transform/output_data/` (GHG_national_Cornerstone_<year>) directly
-        # so the year-Y diagnostics get year-Y GHG data.
-        fbs = _load_cornerstone_ghg_fbs_from_gcs(year)
+        if usa.new_ghg_method and usa.implement_electricity_disaggregation:
+            try:
+                from flowsa import getFlowBySector as get_flowsa_fbs  # noqa: PLC0415
```
WesIngwersen, line 363: flowsa is no longer a separate required package. 

This method must depend on eGRID.  Coding in a fallback to the GHGI is overcomplicated.

WesIngwersen: There are correct references for getting an FBS elsewhere in this module. 
`getFlowBySector `is already imported from 
`bedrock.transform.flowbysector`

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -121,6 +121,24 @@
     get_bea_v2017_summary_to_cornerstone_corresp_df,
 )
 
+
+def _cornerstone_aq_matrix_set(
```
WesIngwersen, line 133: Why is this new internal function in here that is not specific to electricity disaggregation? Can it be avoided?

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -189,13 +206,15 @@ def derive_cornerstone_V(
 
     if apply_inflation:
         V = inflate_cornerstone_V_with_industry_pi(V, target_year=target_year)
+    validate_cornerstone(V, "V")
```
WesIngwersen, line 217: Why add this validation check to this PR?

bedrock/transform/eeio/derived_cornerstone.py
```diff
@@ -244,23 +262,25 @@ def derive_cornerstone_x_after_redefinition(year: int = 0) -> pd.Series[float]:
         iot_before_or_after_redefinition=cfg.iot_before_or_after_redefinition,
     )
     x_cs = expand_vector(x_bea, CS_INDUSTRY_LIST, cs_industry_to_bea_map())
-    x_cs.index.name = 'sector'
-    return _distribute_waste_parent_x_using_v_row_shares(x_cs)
+    x_cs.index.name = "sector"
+    x_out = _distribute_waste_parent_x_using_v_row_shares(x_cs)
+    validate_cornerstone(x_out, "X")
```
WesIngwersen, line 288: again this seems beyond the scope of this PR

bedrock/utils/config/usa_config.py
WesIngwersen: I would suggest keeping only `implement_electricity_disaggregation`

bedrock/utils/taxonomy/bea_v2017_to_ceda_v7_helpers.py
```diff
@@ -3,6 +3,12 @@
 
 import pandas as pd
```
WesIngwersen, line 4: We should not be using this file any more for a main pipeline. I recommend moving all this code into `usa_taxonomy_correspondence_helpers.py `

bedrock/transform/eeio/electricity_disaggregation.py
```diff
@@ -1,16 +1,66 @@
...
+
+GENERATION_GO_SECTOR_NAMES: tuple[str, ...] = (
```
WesIngwersen, line 36: These tuples defining names should be in taxonomy

bedrock/transform/eeio/electricity_disaggregation.py
```diff
@@ -178,3 +236,506 @@ def assert_221100_make_sparsity(V: pd.DataFrame, *, atol: float = 1.0) -> None:
...
+        {
+            "221110": gen_total / total,
+            "221121": trans / total,
+            "221122": dist / total,
+        },
+        dtype=float,
+    )
+
+
+def _diagonal_intersection_weights(w: pd.Series[float]) -> pd.DataFrame:
```
WesIngwersen, line 306: Unclear on what the` w` parameter is for..how do you prepare that. No documentation of that.

bedrock/transform/eeio/electricity_disaggregation.py
```diff
@@ -178,3 +236,506 @@ def assert_221100_make_sparsity(V: pd.DataFrame, *, atol: float = 1.0) -> None:
...
+    data = np.diag([float(w[s]) for s in ELECTRICITY_DISAGG_SECTORS])
+    return pd.DataFrame(
+        data,
+        index=ELECTRICITY_DISAGG_SECTORS,
+        columns=ELECTRICITY_DISAGG_SECTORS,
+        dtype=float,
+    )
+
+
+def build_electricity_disagg_weights(w: pd.Series[float]) -> DisaggWeights:
```
WesIngwersen, line 321: Unclear on what the` w` parameter is for..how do you prepare that. No documentation of that.

WesIngwersen:
Code runs to generate model. Model is generated and `test_electricity_disaggregation` test pass.

bedrock/utils/economic/inflation_helpers_cornerstone.py
jvendries: Leaving electricity mapping separate from  waste parent-fill path in this code block. 

Waste children are already in the BEA to Cornerstone crosswalk, so `_cornerstone_to_ceda_v7_parent() `covers them. 

The disaggregated electricity sectors are not in that map, and with electricity disagg we also drop 221100 and reindex to the 407-sector schema which is something waste doesn’t need here. Keeping that behind its own `if` makes the electricity path clearer than folding inheritance into the waste loop and then reshaping afterward.

bedrock/utils/economic/inflation_helpers_cornerstone.py
jvendries: Removed `price_ratio = price_ratio.reindex(A.index, fill_value=1.0)`.

`get_cornerstone_industry_price_ratio` already returns the active Cornerstone commodity index (including `CORNERSTONE_COMMODITIES_ELEC` when electricity disagg is on), and validated A uses that same ordered index, so the extra reindex was redundant as you point out. This also matches how this helper is structured on main.

bedrock/utils/config/usa_config.py
jvendries: Agreed the end state should be a single flag. Leaving `implement_electricity_reallocation` in the code for now so we can still run step-wise diagnostics only while we settle the final 221100→G/T/D approach. I’ll remove the separate realloc flag once that implementation is determined and we are confortable with it.

bedrock/utils/config/__tests__/test_usa_config.py
jvendries: Removed `test_electricity_disaggregation_config_parsing`. 

Keeping `test_electricity_disaggregation_requires_reallocation_and_waste` and `test_electricity_disagg_requires_waste` until we remove the individual flags

bedrock/transform/eeio/electricity_disaggregation.py
jvendries: Replaced the hardcoded `IO_ACCOUNT_YEAR = 2017` with `usa_base_io_data_year` from the model config.

bedrock/transform/eeio/cornerstone_disagg_pipeline.py
jvendries: Same as on the config flag comment: leaving the two gates for now so we can still run reallocation without the G/T/D split. Once we decisde on the final 3-way split approach and drop the reallocation flag this should become a single gate under `implement_electricity_disaggregation`.

bedrock/transform/allocation/derived.py
jvendries: Deferred to https://github.com/cornerstone-data/bedrock/pull/490#discussion_r3677977274

bedrock/transform/allocation/derived.py
jvendries: Deferred to https://github.com/cornerstone-data/bedrock/pull/490#discussion_r3677977274

bedrock/transform/eeio/derived_cornerstone.py
jvendries: This function is needed for the dual 405/407 schema. It's used on every Cornerstone `Aq` return (i.e., also baseline, not just electricity disagg), so this helper avoids repeating validate + mypy cast across the `Aq` return sites and avoids breaking `CornerstoneAMatrix` Pandera checks under electricity disagg due to the schema change.

Moving it into, e.g., `electricity_disaggregation.py` would be worse as non-disagg code would import from the electricity module and invert the dependency. 

Leaving as is.

bedrock/transform/eeio/derived_cornerstone.py
jvendries: Same reason as for [_cornerstone_aq_matrix_set](https://github.com/cornerstone-data/bedrock/pull/462#discussion_r3678291533): the previous `@pa.check_output(CornerstoneVMatrix)` / `CornerstoneXVectorSchema` (and siblings) assume the fixed 405-sector Cornerstone schema and break once electricity disagg expands to 407. This PR removes those decorators and uses `validate_cornerstone`, which selects 405 vs 407 from config, so we keep schema checks at the derive boundaries (`V`, `x`, etc.) under both modes.

bedrock/transform/eeio/derived_cornerstone.py
jvendries: Same as previous comment

bedrock/utils/taxonomy/bea_v2017_to_ceda_v7_helpers.py
jvendries: Agreed that this file shouldn’t keep growing as the main Cornerstone correspondence path. That said, on main the Cornerstone pipeline still imports from here (`derived_cornerstone`, `year scaling`, `inflation`), so relocating it is a broader refactor than this PR. I’d rather leave the electricity expansion here for now and move the code in a dedicated PR.

bedrock/transform/eeio/electricity_disaggregation.py
jvendries: To be implemented in the PR at the top of the stack to avoid conflicts https://github.com/cornerstone-data/bedrock/pull/508#issuecomment-5124019724

bedrock/transform/eeio/electricity_disaggregation.py
jvendries: Good catch. `w` is the G/T/D BEA GO share series from `build_electricity_disagg_go_weights()` (index 221110/221121/221122, sums to 1). Documented that on both helpers.

bedrock/transform/eeio/electricity_disaggregation.py
jvendries: See reply on previous comment

bl-young (approved):
acknowledging that there are further changes later in the stack - I think this is fine to pull in

jvendries:
### Merge activity

* **Jul 30, 4:20 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/462).
* **Jul 30, 4:20 PM UTC**: @jvendries merged this pull request with [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/462).

PR #458, opened 2026-06-05
feat(eeio): derive y_nab from Adom and scaled_q

bl-young:
cc:
Closes: #456

## What changed? Why?

`y_nab_USA` for the Cornerstone full model (`2025_usa_cornerstone_full_model`) is derived from the same `derive_cornerstone_Aq_scaled()` object that produces snapshotted `scaled_q_USA` and `Adom_USA`, via `backcompute_y_from_q_and_Aq(Adom, scaled_q)`. That enforces IO row balance `q = Adom @ diag(q) + y_nab` and domestic Leontief consistency `q ≈ L_dom @ y_nab`.

The prior path built `y_nab` from BEA Y/trade disaggregation and inflation — parallel to the A/`q` pipeline and inconsistent with snapshotted `scaled_q` (323 commodity sectors failed `q ≈ L·y` at 1% tolerance on the domestic branch).

**`derive_cornerstone_y_nab`** — replaces the BEA disaggregation/inflation body with a three-line backcompute from `derive_cornerstone_Aq_scaled()`. Negative `y_nab` values are retained; `handle_negative_vector_values` breaks the domestic Leontief identity because `L_dom` is dense (24 clipped sectors caused 26 test failures during development).

**`test_compare_output_and_L_y`** — adds a passing Cornerstone domestic case (`Commodity`, `use_domestic=True`). Total Leontief cases (Cornerstone and CEDA) remain `xfail`; Cornerstone total still pairs scaled `Adom + Aimp` with BEA-derived `y` from `derive_cornerstone_Y_and_trade_scaled`.

**Unchanged:** `scaled_q`, `Adom`, `ytot`, `ydom`, `yimp`, `exports` derivation paths. CEDA v7 legacy path. `derive_y_for_national_accounting_balance_usa` router (Cornerstone branch still delegates to `derive_cornerstone_y_nab`).

### Follow-ups (not in this branch)

- **Snapshot bump** — `y_nab_USA` values change; `test_y_nab_usa_snapshot` fails until Phase A regeneration (GCS upload, `.SNAPSHOT_KEY`, `releases.py`, `usa_config.py` Literal). Required before merge if integration tests gate on snapshots.
[...]

PR #470, opened 2026-06-17
fix(validation): atol floor for zero-q Leontief diagnostics

bl-young:
cc:
Closes:

## What changed? Why?

After #458, `test_compare_output_and_L_y[Commodity-True-cornerstone]` failed on CI with sector `S00402` and `max_rel_diff=inf`. The model math was fine; the **comparison metric** was not.

### Why it was failing

The test checks that commodity output matches the Leontief identity `q ≈ L_dom @ y_nab`. It does that by comparing `q` to `L @ y` sector by sector.

`S00402` (“Used and secondhand goods”) is a special commodity with **zero output** (`scaled_q = 0`). After #458, its `y_nab` is a large negative value (~−38B) backcomputed from row balance — that is expected and keeps `q` consistent with `Adom` and `scaled_q`.

On Linux CI, floating-point math in `L @ y` can leave a **tiny nonzero residual** at `S00402` (on the order of 10⁻⁶) while `q` is exactly zero. The old check computed:

```
(q − L@y) / q
```

Dividing by zero produced **`inf`**, which `fillna` did not clear (only `NaN` was handled). The sector was flagged as a failure even though the absolute gap was negligible. The failure was sensitive to platform/BLAS — the same test could pass locally when the residual cancelled to exactly zero.

### What we changed

`validate_result` (used by the Leontief diagnostic and other EEIO checks) now uses the same rule as `numpy.isclose`:

```
|q − L@y| ≤ tolerance × |q| + atol
```

- Sectors with normal output still use **relative** tolerance (1%).
- Sectors with `q ≈ 0` use an **absolute** floor (`atol = 1e-4`) instead of dividing by zero.

`max_rel_diff` is reported as `|diff| / allowed` so values ≤ 1.0 mean pass.

Added unit tests in `TestValidateResult` for the zero-output case, a large-residual failure, and unchanged relative behavior on nonzero sectors.

## Testing

```powershell
uv run pytest bedrock/utils/validation/__tests__/test_eeio_diagnostics.py::TestValidateResult -v
uv run pytest "bedrock/utils/validation/__tests__/test_eeio_diagnostics.py::test_compare_output_and_L_y[Commodity-True-cornerstone]" -m eeio_integration -v
```

### DIFF
--- a/bedrock/analysis/comp_make_use/compare_make_use_by_stages.py
+++ b/bedrock/analysis/comp_make_use/compare_make_use_by_stages.py
@@ -202,8 +202,9 @@ def write_results_csv(
 def _print_result(output: str, result: DiagnosticResult) -> None:
     status = "PASS" if result.passed else "FAIL"
     print(
-        f"{output} (x_V vs Use-side output): max_rel_diff={result.max_rel_diff:.4g}, "
-        f"failures>{result.tolerance:.0%}={len(result.failing_sectors)} — {status}"
+        f"{output} (x_V vs Use-side output): max normalized residual="
+        f"{result.max_rel_diff:.4g} (pass if <= 1.0, rtol={result.tolerance:.0%}), "
+        f"failing sectors={len(result.failing_sectors)} — {status}"
     )
     if not result.passed:
         for sector in result.failing_sectors:
--- a/bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
+++ b/bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
@@ -30,6 +30,7 @@ from bedrock.utils.validation.eeio_diagnostics import (
     compare_output_vs_leontief_x_demand,
     format_diagnostic_result,
     run_all_diagnostics,
+    validate_result,
 )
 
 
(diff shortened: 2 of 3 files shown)


jvendries:
**Documentation follow-up (normalized [`max_rel_diff`](https://github.com/cornerstone-data/bedrock/pull/470/changes#diff-ef63f82534a3af4a6038eee757e679c5f8f387b96e5378b37cd55a781aa35eccR243-R248))**

`validate_result` now reports `max_rel_diff = max(|diff| / (rtol·|value| + atol))` (pass ≤ 1.0), but several docs/log strings still say "relative difference" and imply `max_rel_diff` is comparable to `tolerance`. For example, `DiagnosticResult` [docstring](https://github.com/cornerstone-data/bedrock/pull/470/changes#diff-ef63f82534a3af4a6038eee757e679c5f8f387b96e5378b37cd55a781aa35eccR39) for `max_rel_diff` says `max_rel_diff: Maximum relative difference observed`. It should specify it is not strictly a maximum relative difference to avoid confusion in interpretation of results. As an example, the previously mentioned docstring should say something like:

"Worst-case normalized residual from ``validate_result``:     ``max(|diff| / (tolerance * |value| + atol))`` over sectors. Values
    ``<= 1.0`` are within tolerance; ``> 1.0`` fail. Not comparable to  ``tolerance`` directly. Structural errors may set this to ``inf``. "

Places to update docstrings/comments:

- `DiagnosticResult` docstring: `max_rel_diff` + `tolerance` field descriptions
- `format_diagnostic_result`: log label + doctest example
- `run_all_diagnostics`: failure `RuntimeError` message
- `validate_result`: `tolerance`, `include_details`, Returns sections

- `compare_commodity_output_to_domestics_use_plus_exports`: `tolerance` param doc (also fix stated default 0.05 → 0.01)
- `compare_output_vs_leontief_x_demand`: `tolerance` / `include_details` / Returns
- `commodity_industry_output_cpi_consistency` + `compare_output_from_make_and_use`: note delegation to `validate_result`
- `compare_make_use_by_stages.py`: `_print_result` message (currently `failures>{tolerance:.0%}` is misleading)

**Tests**
- Update `TestFormatDiagnosticResult` if log label changes
- Clarify comments in `TestDiagnosticResult` vs `TestValidateResult` (different scales)
- Refresh integration test comment at line ~364

jvendries (approved):
Tested locally without issues. 

Left a docstring clean up comment [here](https://github.com/cornerstone-data/bedrock/pull/470#issuecomment-4731829348)

PR #472, opened 2026-06-17
fix(ghg): drop FK 5-1-12 flow from UMD 2024 GHGIA path

MoLi7:
cc:
Closes:

## What changed? Why?

Drop the `FK 5-1-12` flow in `load_E_from_flowsa` (`bedrock/transform/allocation/derived.py`) before the gas-collapse step. `FK 5-1-12` is a fluoroketone cover gas from UMD GHGIA Table 4-52 (Magnesium Production) that UMD itself excludes from totals (footnote a). It has no GWP in `GWP100_AR6_CEDA` and falls outside CEDA's Kyoto 7-gas scheme, so it survived the species→7-gas `reverse` map as a stray index entry and broke the `CornerstoneBMatrix` schema (`isin(['CO2','CH4','N2O','HFCs','PFCs','SF6','NF3'])`) in `derive_cornerstone_B_via_vnorm`.

This blocked every 2024-GHG config (`usa_ghg_data_year=2024`: the `umd_2024_ghgia`, `2024_io_ghg`, and FINAL `v0_3` configs); the 2023 path is unaffected since that FBS never carries this flow. The fix is consumer-side, so the pre-built 2024 FBS on GCS stays valid (no regeneration needed).

## Testing

Loaded E for `2025_usa_cornerstone_full_model_v0_3_umd_2024_ghgia`: gas index is now exactly the 7 CEDA gases, no `FK 5-1-12`, no NaNs. Re-running EF diagnostics for the 3 affected configs from this branch.

### DIFF
--- a/bedrock/transform/allocation/derived.py
+++ b/bedrock/transform/allocation/derived.py
@@ -507,6 +507,13 @@ def load_E_from_flowsa() -> pd.DataFrame:
         'CH4_non_fossil'
     )
 
+    # Drop FK 5-1-12 (fluoroketone cover gas, UMD GHGIA Table 4-52 Magnesium
+    # Production). UMD itself excludes it from totals (footnote a), it has no GWP
+    # in GWP100_AR6_CEDA (→ NaN CO2e), and it falls outside CEDA's Kyoto 7-gas
+    # scheme, so the gas-collapse below would leave it as a stray index entry and
+    # break the CornerstoneBMatrix schema. ~6.5e-6 MMT CO2e across 3 sectors.
+    fbs = fbs[fbs['Flowable'] != 'FK 5-1-12']
+
     # Convert values to CO2e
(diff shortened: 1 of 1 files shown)


MoLi7:
* **#472** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/472?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/472?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

MoLi7:
merging this to unblock v0.3-beta snapshot generation, can revisit this along with @bl-young 's review on the new FBS method and make changes if necessary.

bl-young:
Sounds good - I would propose we instead exclude this from the method itself which we should be able to do fairly easily.

PR #475, opened 2026-06-18
468 use equivalent q for a b y nab in scaling method

WesIngwersen:
cc: @MoLi7 
Closes: #468

## Summary

**UPDATED 7/15/2026** to apply config to test off v0.3 model

Extends the scaled-year A-matrix methodology to also produce a consistent scaled B matrix (`E/x @ Vnorm`) using equivalent summary-table scaling for both `x` and `Vnorm`.

### New config flag: `use_scaled_x_and_scaled_Vnorm_for_B`

- Added `use_scaled_x_and_scaled_Vnorm_for_B: bool` to `USAConfig` with a validator that enforces `scale_a_matrix_with_summary_tables` or `scale_a_matrix_with_ceda_method_as_fallback`, and `use_E_data_year_for_x_in_B`, as prerequisites.
- Added config `2025_usa_cornerstone_v0_3_scaling_for_A_and_B.yaml` to exercise the full scaled A+B pipeline.

### `scale_cornerstone_x` and `scale_cornerstone_V`

- Exposed `derive_summary_x_usa` as a public function (was private).
- Added `scale_cornerstone_x` in `cornerstone_year_scaling.py`: scales detail industry gross output using summary x ratios, with optional dollar-year rebasing via the summary industry price ratio when `adjust_summary_dollar_year_before_scaling` is set.
- Added `scale_cornerstone_V` in `cornerstone_year_scaling.py`: scales the detail Make table using summary V block ratios, following the same pattern as `scale_cornerstone_A` but without the out-of-bounds clamping assertion (V row sums are not constrained to ≤ 1).

### Summary-level industry price index

- Added `_aggregate_industry_pi`: x-weighted aggregation of Cornerstone-detail industry PI to BEA summary level using the `bea_v2017_industry__bea_v2017_summary` mapping.
- Added `_get_summary_industry_price_index` (cached): aggregates detail PI to summary using 2017 base-year Cornerstone x as weights (avoids a circular dependency through `scale_cornerstone_x`).
- Added `get_summary_industry_price_ratio` (cached): cross-year ratio of summary industry PI, used as the dollar-year deflator for `adjust_summary_V_dollar_year` and `adjust_summary_x_dollar_year`.
- Added `adjust_summary_V_dollar_year`: rebases summary V from one year's prices to another by dividing rows by the industry price ratio (row-axis only, consistent with `inflate_cornerstone_V_with_industry_pi`).
- Added `adjust_summary_x_dollar_year`: rebases summary x by dividing by the industry price ratio.

[...]

PR #482, opened 2026-06-22
Reconciling data years

WesIngwersen:
cc: @MoLi7 
Closes: http://github.com/cornerstone-data/team-in/issues/43

### Four model specifications
| Model | A matrix | B matrix / x | q |
|---|---|---|---|
| **1** | 2017 USEEIO (fixed) | 2017 E and V-sum x (fixed) | 2017 V column sums |
| **2** | 2017 USEEIO (fixed) | Annual E with BEA time-series x | Authoritative x at `model_base_year=year` |
| **3a** | 2017 USEEIO (fixed) | Annual E with GHG-year x deflated to 2017 | GHG-year x deflated to 2017 |
| **3b** | Commodity-PI inflated to `year` | Annual E with BEA time-series x | Authoritative x at `model_base_year=year` |
| **4** | Summary-table scaled + commodity-PI inflated | Annual E; B scaled 2017→`year` via summary tables | Summary-scaled q |

### `analysis.py` — `main()` loop
For each model × year:
- Clears all function caches and resets config per iteration.
- Derives B, A, q, e_c, L, d, and n in current-year dollars.
- Deflates d and n to `LATEST_TARGET_YEAR` (2024) USD for cross-year comparison.

### Output statistics
- **`n_stats.csv`** — mean and median of `n` across sectors, per model per year.
- **`q_ec_correlation.csv`** — YoY Pearson r (Δq vs Δe_c) and `pct_same_sign` (fraction of sectors where q and e_c move in the same direction) per model per year-transition.
- **`q_ec_correlation_summary.csv`** — wide-format pivot of the above with per-model averages across all transitions.
- **`efs.csv`** — long-format record of d, n, q, e_c per model × year × sector.

### Figures
- **`trends_d_q_ec.png`** — time-series of e_c, q, d, and n indexed to first year = 100, one row per model (including model 1), y-axis shared within each column for direct cross-model comparison.
- **`n_yoy_distribution.png`** — signed YoY violin plot of `n` across models.

### B-matrix scaling guard (`derive_cornerstone_B_non_finetuned`)
`scale_cornerstone_B` is now only invoked when `use_scaled_x_and_scaled_Vnorm_for_B=True`, preventing unintended B scaling in configs that do not opt into it.


MoLi7 (changes requested):
@WesIngwersen this is huge PR to review, so I asked Claude to do it and diff the code changes against what's presented in the latest version of the paper (on Zenodo). Below is its review, lmk what you think

**TL;DR — PR #482 (Reconciling Data Years): one real bug, easy fix, paper conclusions hold**

The bug (Model 4 only): derive_cornerstone_V() is already waste-disaggregated, so compute_x(V) already gives each waste child its correct output. The PR adds _distribute_waste_parent_x_using_v_row_shares(...) to derive_cornerstone_x() — but that helper is only meant for the expand_vector path (where the parent total is duplicated onto every child). Applied to already-split row sums it collapses the 7-child waste total (113.9B → 48.1B), and the scaled derive_cornerstone_x_after_redefinition path then distributes a second time (→ 29.3B). Net: Model 4's waste-sector x is understated ~65%.

Fix (2 lines, isolated): revert derive_cornerstone_x() to plain compute_x(V=...), and in the use_scaled_x_and_scaled_Vnorm_for_B branch return the scaled x without re-distributing. Waste x then correctly sums to 113.9B → 164.4B after scaling. Targeted tests pass (61).

Impact on the paper (I re-ran all 5 models × 7 years, before/after):
- ✅ Instability finding is safe — Model 4 stays the clear outlier; its YoY spread even rises slightly (std 13.4% → 14.5%). Key finding no. 3 holds/strengthens.
- ✅ Models 2/3a/3b are byte-identical before/after, and their Table 2 correlations match the paper exactly. Figure 1's five narrative sectors move <5% and the top-5 set is unchanged — Fig 1 needs no regen.
- ⚠️ Regenerate Model-4 (and Model-1) absolute numbers: mean n drops 0.61 → 0.41 for M4 (and 0.48 → 0.43 for M1, which also uses derive_cornerstone_x); Table 2's Model-4 column shifts (e.g. 2017-18 −0.216 → −0.258); Fig 2 / n_stats.csv / efs.csv for M4 change. So re-run before citing M4/M1 levels.

Also worth a quick look (not blocking the paper, but in the PR):
1. The adjust_summary_A_and_q_dollar_year → adjust_summary_dollar_year_before_scaling rename missed 3 configs (v03_waterfall_g2_methods/g3_data/final.yaml); since USAConfig ignores unknown keys, those silently lose the flag. (Adding extra='forbid' would catch this class of bug.)
2. The B-matrix guard flip (use_E_data_year_for_x_in_B → use_scaled_x_and_scaled_Vnorm_for_B) and cornerstone_industry_avg_margins default True→False change behavior for ~20 other configs — confirm intended + regenerate snapshots.
3. equations_annex.md has README text pasted into the middle of the LaTeX (looks like a merge slip).

WesIngwersen:
Thank you @MoLi7 this is helpful. I'm going to revert the config flag `adjust_summary_A_and_q_dollar_year` name change so its not part of this PR (it overcomplicates it and isn't necessary). I will implement the bug fix affecting model 4 which is a big deal and update the paper with the revised result and plan to release it as a revision on Zenodo.

WesIngwersen:
Here is a before and after the bug fix of the violin plot. It is barely perceptible. 
before
n_yoy_distribution
after
n_yoy_distribution

The change to the e; q; d; n trend plot is imperceptible (since its only select sectors)

FYI the use of the `use_scaled_x_and_scaled_Vnorm_for_B` flag is applied here to model 4 but will be reviewed in PR #475 , so I would reserve comments for that PR for any implications of that flag.  

@MoLi7 this is ready again for review

MoLi7 (approved):
Thank you for the fix - LGTM!

PR #484, opened 2026-06-23
feat(config): point canonical pipeline at 2025_usa_cornerstone_v0_3

bl-young:
cc:
Closes:

## What changed? Why?

The bedrock pipeline default config becomes `2025_usa_cornerstone_v0_3` instead of `2025_usa_cornerstone_full_model`.

- `CANONICAL_USA_CONFIG` in `usa_config.py` is the single source of truth for the config stem.
- `get_usa_config()`, `generate_snapshots`, `generate_diagnostics`, and the `generate_snapshots` / `generate_diagnostics` GitHub workflows use that constant.
- `test_excel_vs_snapshot` and README / snapshots / publish docs reference the v0.3 config as canonical.
- `2025_usa_cornerstone_full_model.yaml` remains available for the v0.2 methodology stack.

## Testing

```powershell
uv run black --check .
uv run ruff check .
uv run pytest
```

### IMPORTANT
Integration tests will fail upon pulling this in until the snapshot is regenerated on main and `.SNAPSHOT_KEY` is updated

### DIFF
--- a/.github/workflows/generate_diagnostics.yml
+++ b/.github/workflows/generate_diagnostics.yml
@@ -7,7 +7,7 @@ on:
         description: "USA config name (without .yaml extension)"
         required: true
         type: string
-        default: "2025_usa_cornerstone_full_model"
+        default: "2025_usa_cornerstone_v0_3"
       sheet_id:
         description: "Google Sheets ID for EF diagnostics"
         required: true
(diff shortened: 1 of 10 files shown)


bl-young:
We may want to consider renaming `2025_usa_cornerstone_full_model` to reflect the v0.2 status of it.

MoLi7 (approved):

PR #491, opened 2026-06-30
config: add v0.3 MECS step on the v0.2 footing

briantobin-99:
## Summary

Adds `2025_usa_cornerstone_full_model_v0_3_ghgi_mecs.yaml` so the MECS step sits on the same footing as the rest of the v0.3 diagnostics chain.

The previous MECS config (`2025_usa_cornerstone_ghg_mecs.yaml`) was fine as a standalone atomic test but two settings made it non-comparable to the other v0.3 steps:

1. It omitted the footing flags `use_E_data_year_for_x_in_B` and `implement_waste_disaggregation`.
2. It pinned the `fbs_schema` snapshot as its baseline instead of the default `v0`.

Both are corrected here. (`update_mecs_method` replaces `new_ghg_method` rather than being set alongside it, since the GHG-method selector is first-match-wins.) The original config is left untouched.

## Testing

- Dispatched `generate_diagnostics` from this branch against both CEDA and USEEIO baselines; both runs succeeded.
- `config_summary` confirms the intended flags and the `v0` baseline snapshot.

### DIFF
--- a/bedrock/utils/config/configs/2025_usa_cornerstone_full_model_v0_3_ghgi_mecs.yaml
+++ b/bedrock/utils/config/configs/2025_usa_cornerstone_full_model_v0_3_ghgi_mecs.yaml
@@ -1,0 +1,5 @@
+use_cornerstone_2026_model_schema: True
+load_E_from_flowsa: true
+update_mecs_method: true
+use_E_data_year_for_x_in_B: True
+implement_waste_disaggregation: True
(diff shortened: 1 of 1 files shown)


bl-young (changes requested):
I think it'd be helpful to include a short narrative like many of the other configs have at the top of this file to describe what it implements.

bl-young (approved):

PR #493, opened 2026-07-06
Add A_margin in prep for Margin EF calc

### LINKED ISSUE #434: Add margin impact into PUR price conversion along with associated config flag
Base it on this useeior function to get an A_margin matrix that can be used in impact calculation.

https://github.com/cornerstone-data/useeior/blob/master/R/CalculationFunctions.R#L487-L524

WesIngwersen:
Progress on: #434 

## Summary

Added `derive_cornerstone_A_margin()` in `bedrock/transform/eeio/derived_cornerstone.py`, a Python port of useeior's [`calculateMarginSectorImpacts`](https://github.com/cornerstone-data/useeior/blob/90f4d7a022a3ecb2043535c30b473cd1068078a8/R/CalculationFunctions.R#L487-L518) `A_margin` construction:

- Uses `derive_margins_cornerstone_usa()` as `model$Margins` and `derive_cornerstone_q()` as `model$q`.
- Added `MARGIN_TYPE_TO_BEA_SECTOR_CODE` and `load_margin_type_to_cornerstone_commodity()` to `bedrock/utils/taxonomy/mappings/bea_v2017_sector__cornerstone_commodity.py` — maps Transportation/Wholesale/Retail to their Cornerstone commodity members via the existing BEA-Sector crosswalk (48TW/42/44RT).
- `_margin_sector_commodity_output_ratio()` is the analog of `calculateOutputRatio(...)$toSectorRatio`, using q-share within each margin group.
- Verified against the real pipeline: 405×405 matrix, 29 nonzero margin-provider rows, no negatives, and column sums match the margin-coefficient totals to float precision.

## Testing
_Show that A has correct shape and has positive values for margins sectors_
```
from bedrock.transform.eeio.derived_cornerstone import derive_cornerstone_A_margin
>A = derive_cornerstone_A_margin()
>print(A.shape)
>print('row sums (margin providers) sample:')
>nz = A[(A.sum(axis=1) != 0)]
>print(nz.shape)
>print(A.sum().sum())
>print((A < 0).any().any())

handling duplicated codes ['221100', '4B0000']
2026-07-06 12:00:16 WARNING  Duplicate sector codes in gross output; aggregating by sum.
2026-07-06 12:00:17 WARNING  Duplicate sector codes in gross output; aggregating by sum.
(405, 405)
row sums (margin providers) sample:
(29, 405)
86.932895721939
False

```

_Verify column sums of A_margin match margin coefficient totals_
```
import numpy as np
from bedrock.transform.eeio.derived_cornerstone import derive_cornerstone_A_margin
from bedrock.transform.iot.derive_PRO_to_PUR_ratio import derive_margins_cornerstone_usa
A = derive_cornerstone_A_margin()
margins = derive_margins_cornerstone_usa()
mc = margins[['Transportation','Wholesale','Retail']].div(margins[\"Producers' Value\"], axis=0).replace([np.inf,-np.inf,np.nan],0.0)
expected_col_sum = mc.sum(axis=1)
actual_col_sum = A.sum(axis=0)
diff = (expected_col_sum - actual_col_sum).abs()
print('max diff (should be ~0 except where a margin group total q is 0):', diff.max())
print('num mismatched beyond 1e-9:', (diff > 1e-9).sum())
print(diff.sort_values(ascending=False).head(10))

max diff (should be ~0 except where a margin group total q is 0): 2.220446049250313e-16
num mismatched beyond 1e-9: 0

```

### DIFF
--- a/bedrock/transform/eeio/derived_cornerstone.py
+++ b/bedrock/transform/eeio/derived_cornerstone.py
@@ -705,7 +705,7 @@ def _margin_sector_commodity_output_ratio(
     """
     ratios: dict[str, float] = {}
     for codes in margin_sector_commodities.values():
-        group_q = q.loc[codes]
+        group_q = q.loc[list(codes)]
         group_total = group_q.sum()
         ratios.update((group_q / group_total).to_dict() if group_total else {})
     return pd.Series(ratios, dtype=float)
@@ -744,7 +744,8 @@ def derive_cornerstone_A_margin() -> pd.DataFrame:
 
     margin_allocation = pd.DataFrame(0.0, index=margin_types, columns=CS_COMMODITY_LIST)
     for margin_type, codes in margin_sector_commodities.items():
-        margin_allocation.loc[margin_type, codes] = output_ratio.loc[codes]
+        code_list = list(codes)
+        margin_allocation.loc[margin_type, code_list] = output_ratio.loc[code_list]
 
     margins_by_sector = margin_coefficients @ margin_allocation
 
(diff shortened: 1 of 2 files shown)


bl-young (approved):
looks good to me - i will follow up on subsequent PR to integrate into the supply chain factor output

PR #519, opened 2026-07-17
adds script to generate `q` needed for ceda figures

bl-young:
cc:
Closes:

## What changed? Why?

This adds the script needed to generate bedrock `q` which is used for the weighted N waterfalls in ceda. ceda will be updated to point to this script upon `FileNotFoundError`.

## Testing

```
uv run python -m bedrock.analysis.v0_3.export_q_v0_3_cornerstone
# optional: --folder-id ... --sheet-title ... --snapshot-key ...
```

### DIFF
--- a/bedrock/analysis/v0_3/README.md
+++ b/bedrock/analysis/v0_3/README.md
@@ -16,11 +16,25 @@ Plot outputs land in `output/release_v0_3/` or `output/release_v0_v03_groups/`
 
 | Script | Purpose |
 |--------|---------|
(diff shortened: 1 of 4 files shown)


MoLi7 (changes requested):
eh.. how about we write `q` to GSheets of EF diagnostics? keeping local copies seems dangerous to me

bl-young:
ok ill swap to sheets and make the update in ceda

MoLi7 (approved):
thank you for changing it so quickly!

bl-young:
Shows up [here](https://docs.google.com/spreadsheets/d/1kq-mDZZiiOk98F5G5nkYvOCGxNjHWBBk_bAf8pyw6AA/edit?usp=drive_link)

Issue #523, opened 2026-07-22
Step 1: Final Uses with Trade


WesIngwersen:
Umbrella for Step 1 of the nowcast. **Rescoped 2026-08-14: this is no longer one Final Demand FBS.** Final uses split into three FBS methods that are built, validated and maintained separately, with trade in scope because exports are a final use and the same extraction serves the Supply-side import columns.

Plan: [`bedrock/analysis/nowcasting/plan.md`](https://github.com/cornerstone-data/bedrock/blob/nowcast/bedrock/analysis/nowcasting/plan.md) §Step 1.

### A. `NIPA_final_consumption` — the NIPA-sourced columns

`F01000` PCE, `F02E00`/`F02N00`/`F02R00`/`F02S00` private fixed investment, and the twelve `F06*`/`F07*`/`F10*` federal/state/local columns. Everything attributable from the NIPA underlying tables against the PCE and PEQ bridges or the detail Use table.

Renamed from `NIPA_FD` — the old name promised a whole final-demand block it never delivered. It has never contained trade or inventories: `F03000` is hardcoded zero and `F04000` is bolted on inside `derive_initial_Y_pur`, so this rename mostly ratifies what the method already is.

- [x] PCE Bridge for allocation
- [x] PEQ Bridge for allocation of investment in equipment
- [x] clean up the NIPA FBAs
- [x] PCE and equipment reproduce their bridges cell for cell (#630, #631)
- [x] government columns reproduce the Use table cell for cell (#633, #634)
- [ ] rename the method and its yamls off `NIPA_FD`
- [ ] review all activity mappings
- [ ] 2018–2024 have no PCE activity sets — #621
- [ ] port the common yaml — #495

### B. Trade — #526, #528

Exports (`F04000`) and the Supply-side import columns from one extraction. `Trade_Exports_2017` and `Trade_Imports_2017` exist and are wired into the nowcast `Y`/`MCIF`. Currently blocked: the trade yamls still use `target_naics_year` where the code reads `target_schema_year` since #630, so they raise `KeyError` and take `derive_initial_Y_pur` down with them.

### C. Change in inventories — #529, #530, #531

`F03000`, the full 402-row column. Not NIPA-total-only: that ships one scalar where the SUT needs a column and leaves the allocation to Step 5's RAS, which has neither a seed nor a sign-safe structure for a column running 3× gross to net across 61 negative commodities.

## Above the three — composition and validation

[...]

PR #524, opened 2026-07-23
523 update nipa fd fbs

### LINKED ISSUE #496: Step 1a: NIPA Private investment in equipment table not used in FBS method
Table 5.5.5U has the data on investment in equipment and an FBA is made for it:

https://github.com/cornerstone-data/flowsa/blob/91fa085d15ef8ed45ee14568c521b5ed8a6b33f6/flowsa/methods/flowbyactivitymethods/BEA_NIPA.yaml#L23

But it is not used in the flow-by-sector method

https://github.com/cornerstone-data/flowsa/blob/91fa085d15ef8ed45ee14568c521b5ed8a6b33f6/flowsa/methods/flowbysectormethods/NIPA_FD_common.yaml

WesIngwersen:
cc:
Closes: #496, #525 
Makes progress on: #523 

## What changed? Why?

`Cornerstone_2025_target.yaml` .  
- Final demand sectors were not being properly mapped from 3 to 6-digit for codes like `F02R` --> `F02R00`. So have to add them as NAICS-6 sectors. 

`NAICS_2017_Codes`. `BEA_Custom_Codes`, `Sector_Levels.csv`,  Add codes for special S* sectors at 2 to 6 digit levels. 

### NIPA FBA
- Strip trailing notes like (30) from name in ActivityProducedBy. Apply revised, cleaner names in the activity to sector crosswalks and activity sets in the FBS method. In a limited number of cases manual changes did have to be applied to names when they were duplicated without the notes. These are handled with separate activity_sets. 
- Strip padding around names in between '/' or ','
- Add proper formatting in BEA_NIPA.py 

### New PEQ Bridge FBA with associated crosswalks 
Parallels existing PCEBridge flow; latter storage location moved to put bridges together
Needed for use in allocating the Investment in Equipment

### NIPA FD FBS (only to 2017 as of now)

- All activitytosector crosswalks updated with NIPA name updates.

- Change target from `BEA_Detail_2017 `to `Cornerstone_2025_target`

- Adds PCE data allocation using PCE bridge. Heavy fixes from original flowsa version.

- Remove `FD_Structures_1` for now which is breaking the code. FD_Structures1_used targeted a NIPA line ("Net purchases of used structures", Residential variant) that genuinely has no sector in the crosswalk — its primary FBA was always empty. That empty dataframe raised a ValueError deep in harmonize_geoscale, caught by a blanket except ValueError several frames up in flowbyactivity.py's activity-set-splitting logic — which discards the entire method's output, not just the one failing activity_set. That's why 20 correctly-computed activity_sets (including the working -1.3% FD_PCE result) were vanishing into a silent 0-row result. Removed that dead activity_set.

## New nowcast.py module
Starts creation of initial Y based on the FBS

## Testing
Diagnostic script added to being comparing initial Y to BEA Supply Use 2017 Detail Y


catherinebirney:
@WesIngwersen all the changes so far look good! Removing the line number from the activity names does clean up the code. Good catch on the sector level errors ... and I can look into the issue behind FD_Structures_1 - probably best to modify how that error drops code rather than commenting out an activity set that triggers it. 

I did go ahead and update additional crosswalks with the BEA codes, not that it impacts your work here

WesIngwersen:
There is now a blocking issue, #533, which is specifically affecting the `FD_PCE`and `FD_Structures*` activity sets in the NIPA_FD method.

FYI @catherinebirney should we deal with that in this branch or branch off it?  This branch has the new crosswalks and the new FBS that is showing this error.

catherinebirney:
> There is now a blocking issue, #533, which is specifically affecting the `FD_PCE`and `FD_Structures*` activity sets in the NIPA_FD method.

@WesIngwersen

Part of the problem is that the [BEA_2017_Detail](https://github.com/cornerstone-data/bedrock/blob/523_update_NIPA_FD_FBS/bedrock/utils/mapping/activitytosectormapping/NAICS_Crosswalk_BEA_2017_Detail.csv#L970) mapping does not retain the newly added housing codes `531HSO`, `531HST`, `531ORE`. They get  mapped to standard NAICS. So the BEA data used to proportionally attribute ends up only mapped to 531 after aggregation. The only other FBS that call on the BEA_2017_Detail data are the CAP_HAP) - so we can change the activity to sector mapping to keep those BEA codes without any negative impacts. 

The other part of the problem...is definitely a bug in how sector_aggregation() is called after we already proportionally attribute. In the final generated FBS, `531HSO`, `531HST`, `531ORE` should all stay as 0 (due to being dropped in the BEA crosswalk). And it is a problem that the FlowAmount for 531 is duplicated/added to those 3 sectors. 

I can look into this issue more tomorrow and think about the fix for sector_aggregation().

WesIngwersen:
> > There is now a blocking issue, #533, which is specifically affecting the `FD_PCE`and `FD_Structures*` activity sets in the NIPA_FD method.
> 
> @WesIngwersen
> 
> Part of the problem is that the [BEA_2017_Detail](https://github.com/cornerstone-data/bedrock/blob/523_update_NIPA_FD_FBS/bedrock/utils/mapping/activitytosectormapping/NAICS_Crosswalk_BEA_2017_Detail.csv#L970) mapping does not retain the newly added housing codes `531HSO`, `531HST`, `531ORE`. They get mapped to standard NAICS. So the BEA data used to proportionally attribute ends up only mapped to 531 after aggregation. The only other FBS that call on the BEA_2017_Detail data are the CAP_HAP) - so we can change the activity to sector mapping to keep those BEA codes without any negative impacts.
> 
> The other part of the problem...is definitely a bug in how sector_aggregation() is called after we already proportionally attribute. In the final generated FBS, `531HSO`, `531HST`, `531ORE` should all stay as 0 (due to being dropped in the BEA crosswalk). And it is a problem that the FlowAmount for 531 is duplicated/added to those 3 sectors.
> 
> I can look into this issue more tomorrow and think about the fix for sector_aggregation().

Root cause: `FlowBySector.sector_aggregation()` (bedrock/transform/flowbysector.py:270) does a plain `merge(naics_key, how='left') `to expand sector codes to the target industry_spec's granularity. For most codes this is a clean 1:1 (or genuine many:1 collapse) relationship. But Cornerstone_2025_target.yaml's NAICS_6 list explicitly splits 531 (generic real estate) into three finer codes — industry_spec_key confirms '531' maps to 4 target rows: 531, 531HSO, 531HST, 531ORE. A left merge against a 1-to-many mapping duplicates the row for each match, carrying the full, undivided FlowAmount into every one — it doesn't know how to split the dollar value, so it just copies it. Any commodity/sector landing on bare 531 upstream (imputed rent lines from FD_PCE and FD_Structures* in our case) gets counted three times over once this step runs.

WesIngwersen:
@catherinebirney This PR is getting too big. I think you can work on that issue in a new branch since it appears it could have wide ranging effects (hopefully not). I recommend we pull this in and then you branch off to work on that issue, and i can branch to keep going on this work.

bl-young:
Do we think this will impact our main GHG FBS? We should run the integration tests to confirm.

catherinebirney:
@WesIngwersen and @bl-young  I fixed the issue locally this morning - I tested the 2021 Cornerstone GHG method and no changes to the results. But I can throw the changes in another branch to double check with the integration tests

catherinebirney (approved):
I didn't look closely at `nowcast.py` or `nowcast_intitial_Y_pur_baseline` - but everything else looks good and extract and transform datasets generate as expected.

PR #532, opened 2026-07-23
Support service-account impersonation in google-login for blocked ADC Drive scopes

briantobin-99:
cc: @jvendries
Closes:

## What changed? Why?

Google is rolling out a block on Drive/Sheets ("sensitive") scopes for gcloud's **default** OAuth client ([docs](https://docs.cloud.google.com/docs/authentication/troubleshoot-adc#access_blocked_when_using_scopes)). For accounts where enforcement has landed, `./scripts/google-login` fails at the browser consent screen with "Access blocked" — re-picking accounts or re-running with different scopes cannot fix it. gcloud's own warning names the remedies: "you must provide your own client ID **or use service account impersonation**."

- `scripts/google-login`: plain login unchanged as the default; adds an opt-in impersonation mode:

  ```bash
  BEDROCK_IMPERSONATE_SA=github-actions-cornerstone@cornerstone-data.iam.gserviceaccount.com \
    ./scripts/google-login
  ```

  Drive/Sheets scopes ride on short-lived tokens minted for the diagnostics service account (the same identity CI uses to write EF diagnostics sheets, already shared on the Cornerstone Drive), so the blocked consent screen never appears — the user consent requests only `cloud-platform`. Requires a one-time `roles/iam.serviceAccountTokenCreator` grant on the service account (project owners can add people with one command).

- `pyproject.toml` / `uv.lock`: bump google-auth floor to `>=2.44.0`. The previously locked 2.43.0 has a bug in impersonated-credential refresh (`TypeError: 'str' object is not callable`, [google-auth #1884](https://github.com/googleapis/google-auth-library-python/issues/1884)), fixed upstream in 2.44.0. Plain user-credential ADC never hits this code path, which is why 2.43.0 worked fine until now.

- Also adds the `spreadsheets` scope to the plain login.

## Testing

- End-to-end from a real workstation: impersonated login (consent URL confirmed to request only `cloud-platform`), Drive folder listing, and blank-sheet creation in the EF diagnostics folder via `bedrock.utils.io.gcp`.
- A/B on the google-auth bump with identical code and credentials: 2.43.0 reproduces the impersonated-refresh `TypeError`; the relocked version works.
- `bash -n scripts/google-login`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,7 @@ These flags are grouped into themes:
 |---|---|
 | Schema / taxonomy | `use_cornerstone_2026_model_schema`, `implement_waste_disaggregation`, `implement_electricity_reallocation` |
 | Economic IOT (input-output tables) | `use_E_data_year_for_x_in_B`, `iot_before_or_after_redefinition` |
-| GHG attribution — sector methods | `update_transportation_ghg_method`, `update_electricity_ghg_method` |
-| GHG attribution — gas coverage | `update_flowsa_refrigerant_method`, `add_new_ghg_activities` |
+| GHG attribution | `new_ghg_method`, `update_mecs_method`, `v0_3_umd_2023_ghgia`, `v0_3_umd_2024_ghgia` |
 | Data vintage | `model_base_year`, `usa_base_io_data_year`, `ipcc_ar_version` |
 
 See [`USAConfig`](bedrock/utils/config/usa_config.py) for the full list.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -22,7 +22,8 @@ These flags are grouped into themes:
 |---|---|
 | Schema / taxonomy | `use_cornerstone_2026_model_schema`, `implement_waste_disaggregation`, `implement_electricity_reallocation` |
 | Economic IOT (input-output tables) | `use_E_data_year_for_x_in_B`, `iot_before_or_after_redefinition` |
-| GHG attribution | `new_ghg_method`, `update_mecs_method`, `v0_3_umd_2023_ghgia`, `v0_3_umd_2024_ghgia` |
+| GHG attribution — sector methods | `update_transportation_ghg_method`, `update_electricity_ghg_method` |
+| GHG attribution — gas coverage | `update_flowsa_refrigerant_method`, `add_new_ghg_activities` |
```
bl-young: I think this readme change got reverted accidentally

uv.lock
```diff
@@ -140,7 +140,7 @@ wheels = [
 
 [[package]]
 name = "bedrock"
-version = "0.3.0"
+version = "0.1.0"
```
jvendries: Why did this change from v0.3.0 to v0.1.0?

README.md
briantobin-99: Thanks for the catch

uv.lock
briantobin-99: It's corrected now. Let me know if this branch takes care of the problems you were having!

jvendries (approved):
Tested successfuly using 

` gcloud auth application-default login --impersonate-service-account=github-actions-cornerstone@cornerstone-data.iam.gserviceaccount.com`

to log in with the impersonation account credentials and then running 

`uv run python -m bedrock.utils.validation.analysis.combine_ef_diagnostics \
  --combo v0_to_v03_useeio_groups \
  --output-xlsx bedrock/analysis/v0_3/output/release_v0_v03_groups/ef_diagnostics_merged_v0_to_v03_useeio_groups_useeio.xlsx`

which is one of the commands I was unable to run before

WesIngwersen:
To get the impersonation to run by default in my environment, in the project in my bash shell i had to run the one time command to set this variable for this project
```
setx BEDROCK_IMPERSONATE_SA "github-actions-cornerstone@cornerstone-data.iam.gserviceaccount.com"
```
The restart my terminal and now I can authenticate and get an ADC when i run `./scripts/google-login`

FYI `setx` is setting this as a user wide (not project scope) variable . This is the same as manually setting it up in Windows via the Environmental Variables\User variables in the Settings

PR #534, opened 2026-07-24
533 sector aggregation

### LINKED ISSUE #533: Flow-by-sector aggregation error in expanding sector codes affecting 531* sectors
to expand sector codes to the target industry_spec's granularity

FlowBySector.sector_aggregation() (bedrock/transform/flowbysector.py:270) does a plain merge(naics_key, how='left') to expand sector codes to the target industry_spec's granularity. For most codes this is a clean 1:1 (or genuine many:1 collapse) relationship. But Cornerstone_2025_target.yaml's NAICS_6 list explicitly splits 531 (generic real estate) into three finer codes — industry_spec_key confirms '531' maps to 4 target rows: 531, 531HSO, 531HST, 531ORE. A left merge against a 1-to-many mapping duplicates the row for each match, carrying the full, undivided FlowAmount into every one — it doesn't know how to split the dollar value, so it just copies it.

catherinebirney:
cc:
Closes: #533 

## What changed? Why?
Addresses the issue identified of #533 where in sector_aggregation() if there is a 1:many mapping, but less aggregated child sectors were missing, an existing more aggregated sector has the values duplicated and added on to the other sectors - when those sectors should have remained 0. This function is modified to meet its original purpose - ensures only sector aggregation occurs. 

Updates the BEA Detail 2017 mapping to keep the three housing BEA sectors `531HSO`, `531HST`, `531ORE` instead of mapping these BEA codes to NAICS

## Testing

[integration test](https://github.com/cornerstone-data/bedrock/actions/runs/30106202999)

I ran `NIPA_FD_2017` for `FD_Structures1: "Brokers' commissions and other ownership transfer costs"` which was previously duplicating a 531 value for the three specific BEA housing codes. That is no longer the case. The FBS results are identical before and after running `sector_aggregation()`

image


catherinebirney:
@bl-young @WesIngwersen integration test passes

WesIngwersen (approved):

PR #541, opened 2026-07-27
Analysis package to quickly compare NIPA or other data to BEA IOT data for reference year

WesIngwersen:
## compare_NIPA_to_IOT: compare a candidate dataset against a BEA reference table

Adds `bedrock/analysis/compare_NIPA_to_IOT`, for checking a dataset against a BEA
input-output table cell by cell and in total, without first building a crosswalk.
The current candidate source is BEA NIPA; the reference is any BEA IOT in bedrock.

14 files, +3,924 lines, all new except a 3-line `pyproject.toml` entry. No existing
behavior changes.

```python
from bedrock.analysis.compare_NIPA_to_IOT import bea_matrix_row, compare, nipa_flat_table

result = compare(
    candidate=nipa_flat_table('T60200D', 2017).leaves(),   # NIPA table 6.2D
    reference=bea_matrix_row('V00100'),                    # Use SUT detail, compensation
    rollup='industry_to_summary',                          # 402 detail -> 71 summary
)
print(result.report())
result.to_csv('out/compensation.csv')
```

### Structure

| module | |
|---|---|
| `series.py` | `LabeledSeries`, a `code / name / value / level` frame, plus `leaves`, `subtree`, `select`, `rollup`, `merge_codes` |
| `loaders.py` | BEA reference matrices and NIPA/FBA/csv candidates |
| `matching.py` | the alignment cascade, one-to-one, with the pass recorded per pair |
| `hierarchy.py` | residual-label conventions per source dialect |
| `compare.py` | `compare()` and the report |

Reference loaders are `bea_matrix_row` and `bea_matrix_column`. Candidate loaders
are `nipa_flat_table`, `nipa_sheet`, `fba_series`, `table_series`, `frame_series`.

### Matching

Passes run in order: overrides, code, crosswalk, name, hierarchy relations, fuzzy
name. Each pair records which pass produced it and a 0–1 score. Matching is
one-to-one, and a key ambiguous on either side is reported rather than resolved.

The fuzzy pass runs by default and only sees what the earlier passes left over. It
is gated twice: a 0.88 `difflib` cutoff, and `token_relation`, which rejects a pair
differing by a substituted content word. "Support activities for mining" and
"Support activities for printing" score 0.90 and are industries a factor of 20
apart, so that pair is refused. Fuzzy pairs are labelled `fuzzy` in `method`, carry
their `score`, and are counted on their own line in the report's `MATCHED BY`
block. `on='code'` or `on='name'` excludes the pass.

On the two worked examples the fuzzy pass adds nothing — the best unmatched
nearest-name score is 0.846, below the cutoff — so the gates, not the pass, are
what determine behavior on this data.

Hierarchy relations are read from labels before any similarity is computed. Where
BEA carves a residual out of a parent and names it after the parent, the pass
records a parent/child relation instead of a match and holds both rows out of the
fuzzy pass.

### Naming BEA matrices
Three axes decide whether two BEA tables are comparable, and each matrix name
states all three. There are no framework-silent aliases: `Use_detail` does not say
which of the two detail Use tables it means, and `Use_MUT_detail` does not say
which redefinition it carries, so neither resolves.

| matrix | framework | valuation | redefinition |
|---|---|---|---|
| `Use_SUT_detail` *(default)* | Supply-Use | purchaser cells, basic + producer totals | — |
| `Supply_SUT_detail` | Supply-Use | basic cells, basic + purchaser totals | — |
| `Use_MUT_detail_after_redef` | Make-Use | producer | after |
| `Make_MUT_detail_after_redef` | Make-Use | producer | after |
| `Import_MUT_detail_after_redef` | Make-Use | producer | after |
| `Use_MUT_detail_before_redef` | Make-Use | producer | before |
| `Make_MUT_detail_before_redef` | Make-Use | producer | before |
| `Import_MUT_detail_before_redef` | Make-Use | producer | before |

The SUT Use table's valuation is not one basis. Its cells are at purchaser value,
and its industry columns are totalled at both basic and producer over those cells,
so `MatrixSpec.valuation` describes the cells and the matrix `note` covers the
totals. `About_BEA_IOT_table_valuation_differences.md` documents the differences
and the identities behind them.

### Reading NIPA

`nipa_flat_table` reads `extract/input_data/BEA_NIPA/FlatFiles.ZIP`, the archive
`BEA_NIPA.py` parses, so a comparison and the FBA it checks stay on one BEA
vintage. It delegates to `BEA_NIPA.flat_files_local_path()` for that path.
Hierarchy comes from `SeriesCodeParents`, narrowed to parents that are also lines
of the requested table.

`nipa_sheet` reads the published workbooks and is kept for one reason: the flat
files carry no footnote block, and footnotes are the only description of NIPA's
residual lines.

### Granularity

`rollup` aggregates the reference before matching rather than switching to BEA's
published summary table. Each cell keeps `n_detail` and `detail_members`, so a cell
that is 1:1 with a single detail code is distinguishable from one summing 29.

### Verification

- 83 unit tests, all synthetic — no GCS, no BEA workbooks. `ruff` and `mypy` clean.
- Two worked examples run end to end:
  - NIPA 6.2D vs Use SUT `V00100`: 69/69 cells, −1 million on $10.4 trillion.
  - NIPA 3.5 vs `T00OTOP` / `T00TOP`: −9 and −13, and a two-cell split against the
    Supply table's `MDTY` / `TOP`.
- Both NIPA loaders produce identical figures across all three stages of the
  compensation example.

### Notes for review

- `pyproject.toml` gains the package; nothing else outside the new directory changes.
- The examples take about a minute each, mostly loading BEA workbooks from GCS.
- Fuzzy matching is on by default, gated by the 0.88 cutoff and `token_relation`.
  Prefer `overrides` for pairs you want stated rather than inferred.


bedrock/analysis/compare_NIPA_to_IOT/__init__.py
```diff
@@ -0,0 +1,90 @@
+"""Nimble comparison of candidate data against BEA reference tables.
+
+Built for reconnaissance, not for provenance: alignment falls back to fuzzy name
```
bl-young, line 3: This makes it seem like fuzzy matching happens automatically, but you have to turn that on correct?

bedrock/analysis/compare_NIPA_to_IOT/README.md
```diff
@@ -0,0 +1,324 @@
+# compare_NIPA_to_IOT
+
+Check a candidate dataset against a BEA reference table — cell by cell and in
+total — without first building an exact crosswalk.
+
+Built for reconnaissance. Matching cascades from codes to names to fuzzy names
```
bl-young: same as the other comment, makes it seem like fuzzy is default

bedrock/analysis/compare_NIPA_to_IOT/loaders.py
```diff
@@ -0,0 +1,838 @@
...
+    ),
+}
+
+#: bedrock's own, framework-silent names, still accepted at call sites
+MATRIX_ALIASES = {
+    'Supply_detail': 'Supply_SUT_detail',
+    'Use_detail': 'Use_MUT_detail',
+    'Make_detail': 'Make_MUT_detail',
+    'Import_detail': 'Import_MUT_detail',
+}
```
bl-young: It feels a bit dangerous to leave the `Use_detail` key here unspecified since that could refer to either SUT or MUT. Is this `MATRIX_ALIASES` even necessary?

bl-young (approved):
I'm not following all the intricacies of the code but I was able to generate the example outputs and the idea is a good one.  See my few minor comments.

bedrock/analysis/compare_NIPA_to_IOT/loaders.py
WesIngwersen: This is replaced with BEA_MATRICES

```
BEA_MATRICES: dict[str, MatrixSpec] = {
    'Use_SUT_detail': MatrixSpec(
        'SUT',
        'PUR',
        note=(
            'cells and final demand are at purchaser value (T019 = Supply T016 '
            '37,094,434 for all 402 commodities); the industry columns are '
            'totalled twice over them, T018 33,772,568 at basic and T005+VAPRO '
            '34,468,127 at producer, so value added comes both ways too, VABAS '
            '18,916,542 and VAPRO 19,612,109'
        ),
    ),
    'Supply_SUT_detail': MatrixSpec(
        'SUT',
        'BAS',
        note=(
            'commodity x industry cells are domestic output at basic value; the '
            'trailing columns then bridge to the other two bases, all 402/402 '
            'per commodity: T013 total supply at basic 36,398,867 (= T007 + MCIF '
            '+ MADJ), + T014 margins (TRADE+TRANS, ~0 in total) + T015 taxes '
            '(MDTY+TOP+SUB, 695,565) = T016 37,094,434 at purchaser. Producer is '
            'not published but derives as T013 + T015 = T016 - T014 = 37,094,432 '
            '-- taxes but not margins. Note SUB is stored negative here and so is '
            'added, where the Use table carries T00SUB positive and subtracts it'
        ),
    ),
    'Use_MUT_detail_after_redef': MatrixSpec('MUT', 'PRO', 'after'),
    'Make_MUT_detail_after_redef': MatrixSpec('MUT', 'PRO', 'after'),
    'Import_MUT_detail_after_redef': MatrixSpec('MUT', 'PRO', 'after'),
    'Use_MUT_detail_before_redef': MatrixSpec('MUT', 'PRO', 'before'),
    'Make_MUT_detail_before_redef': MatrixSpec('MUT', 'PRO', 'before'),
    'Import_MUT_detail_before_redef': MatrixSpec('MUT', 'PRO', 'before'),
}

```

bedrock/analysis/compare_NIPA_to_IOT/__init__.py
WesIngwersen: It's not by default and clarified...but it's lowest priority and if special token words are in match like "Support activities" it will reject the match

PR #542, opened 2026-07-27
Cache BEA_NIPA's FlatFiles.ZIP instead of re-downloading it every run

WesIngwersen:
`BEA_NIPA` parsed BEA's national-accounts archive straight out of `resp.content` via `io.BytesIO`, so every `generateFlowByActivity` refetched ~19MB from `apps.bea.gov` and left nothing behind. It was one of the last extractors with no raw-source cache — the `gcs_fxn` + `extract_data_from_raw_sources` pattern that a dozen other sources use (EPA_GHGI, BLS_QCEW, EIA_MECS, USDA_CoA, BLM_PLS…) was simply never wired up here.

## What changed

`bea_nipa_call` now writes the archive to `extract/input_data/BEA_NIPA/FlatFiles.ZIP` before reading it, and a new `bea_nipa_load_gcs` reads that cache, falling back to `gs://cornerstone-default/extract/input-data/BEA_NIPA/FlatFiles.ZIP` when it is absent. The shape follows `EIA_ElectricPowerAnnual`, which caches its edition zip the same way.

```yaml
call_response_fxn: !script_function:BEA_NIPA bea_nipa_call
gcs_fxn: !script_function:BEA_NIPA bea_nipa_load_gcs
extract_data_from_raw_sources: False
```

`flat_files_local_path()` is exported because `bedrock/analysis/compare_NIPA_to_IOT` reads the same archive: a comparison against a NIPA table and the FBA built from it now stay on one BEA vintage by construction rather than by coincidence.

## Testing

All three ways in were run against the existing 2017 parquet, and **each reproduces it byte for byte** (690 rows × 20 cols, `DataFrame.equals` → `True`, $257.5T):

| path | how it was forced | result |
|---|---|---|
| local cache | `make_url_request` patched to raise | identical — proves nothing reaches the network |
| GCS | local cache removed, same patch | downloads 19,751,423 bytes, md5 identical to the local copy |
| BEA url | `extract_data_from_raw_sources: True`, cache deleted | re-downloads and re-caches; archive md5 unchanged |

The md5 checks matter: `4c235d4ff5e181b7fab0830d6bf5964f` is the same archive locally, in the bucket, and freshly fetched from BEA, so the cache is not pinning a stale vintage.

## Notes for review

- **Behavior change on a cache miss.** With `extract_data_from_raw_sources: False`, a machine with neither the local file nor GCS access no longer silently falls back to BEA — it raises. That is the same contract every other `gcs_fxn` source already has, and the error says to flip the flag. Worth knowing if anyone runs this without bucket credentials.
- **Refreshing after a NIPA release** is: flip the flag to `True`, run once, flip it back, re-upload the archive to the bucket.
- **Follow-up, not in this PR:** `compare_NIPA_to_IOT.loaders.nipa_flat_files_path()` recomputes the same path instead of delegating to `flat_files_local_path()`. That package is on a separate branch off `main`; whichever merges second should collapse the two.


bedrock/extract/bea/BEA_NIPA.py
```diff
@@ -6,32 +6,83 @@
...
-                            'TableId:LineNo': 'Table_and_Line',
-                        }
-                    )
-                )
 
-    return df_list
+def flat_files_local_path(source: str = 'BEA_NIPA') -> str:
+    """Where the flat-file archive is cached: ``extract/input_data/BEA_NIPA/``.
+
+    Also the archive :mod:`bedrock.analysis.compare_NIPA_to_IOT` reads, which is
```
bl-young: `:mod:` ?

bedrock/extract/bea/__tests__/test_bea_nipa.py
bl-young, line 1: I generally haven't been adding tests like this as the proof tends to be in being able to pull the data when we need it. I find it just adds a lot of bloat. But if you feel strongly we can keep it.

bl-young:
Yes this update makes sense, I see you added it to GCS already.

bedrock/extract/bea/__tests__/test_bea_nipa.py
WesIngwersen: Nope we don't need it. i can remove it.

bedrock/extract/bea/BEA_NIPA.py
WesIngwersen: dropped this note

bl-young (approved):

PR #522, opened 2026-07-17
 chore(allocation): remove registry E path and trim dead EPA loaders

bl-young:
cc:
Closes:

## What changed? Why?

Removes the per-source EPA allocation registry used when `load_E_from_flowsa` was false. `derive_E_usa()` always loads E from flowsa FBS. Runtime Cornerstone / v0.3 already used that path; the registry only served the retired CEDA full-model build.

Also trims unused public `load_*` helpers in `bedrock/extract/allocation/epa.py` and matching `epa_constants` table-name maps that only served those registry allocators. Flowsa / MECS / industrial helpers keep the GCS table loader and the remaining public loaders.

### Deleted — registry E

| Surface | Notes |
|---|---|
| `derive_E_usa_by_gas` / `derive_E_usa_emissions_sources` | Registry aggregation |
| `registry.py` / `ALLOCATED_EMISSIONS_REGISTRY` | |
| `derive_emissions_source_B.py` | Unused outside registry |
| `constants.py` (`EmissionsSource`) | Registry-only |
| `waste_utils.py` | Registry waste allocators only |
| Entire `allocation/ch4/` and `allocation/n2o/` | |
| Most of `allocation/co2/` and `allocation/other_gases/` | Registry process / transport / HFC allocators |
| Gas `eeio_integration` tests + `allocation/conftest.py` | |
| Unused `CEDA_EMISSIONS_SOURCES` list | `utils/emissions/ghg.py` |

### Deleted — dead EPA extract loaders

| Surface | Notes |
|---|---|
| Process / gas `load_*` in `epa.py` | `load_ch4_*`, `load_n2o_*`, `load_hfc_*`, `load_pfc_*`, `load_recent_trends_*`, orphan CO2 process loaders, etc. |
| Matching `EPA_TABLE_NAMES` / year→table maps | `epa_constants.py` |
| AST scrape helpers | `return_emissions_source_table_numbers` and related |

### Kept (flowsa / extract consumers)

| Surface | Why |
|---|---|
| `load_E_from_flowsa()` / `_select_flowsa_ghg_method()` | Primary E path |
| `co2/industrial_*` + `co2/non_energy_fuels_*` | `extract/eia/EIA_MECS.py` |
| `other_gases/common_ratios.py` | BEA CEDA foam / refrigeration extractors |
| `transportation_fuel_use/` | BEA transport + non-energy transport |
| `allocation/mappings/v7/` + `mappings/cornerstone/` | Live MECS industry maps |
[...]

PR #543, opened 2026-07-27
Retire legacy CEDA E-matrix allocator path

briantobin-99:
cc:
Closes:

## What changed? Why?

Since #521 no loadable config sets `load_E_from_flowsa: false`, so the registry-driven CEDA allocator path behind `derive_E_usa()` was unreachable dead weight (~6.9k lines / 124 files). This deletes it and makes the flowsa FBS path unconditional:

- Delete `transform/allocation/{registry,constants,waste_utils,derive_emissions_source_B}.py`, all of `ch4/` and `n2o/`, the non-keeper `co2/` and `other_gases/` allocators, `mappings/v5/`, and the four `eeio_integration` allocator test suites plus their `conftest.py`.
- Trim `extract/allocation/epa.py` (1,197 → ~330 lines) to the five loaders still used by FBA builders; delete `extract/allocation/{usda,utils}.py`. All `ipcc_ar_version` branching lived in the deleted loaders — the field is now unread by the pipeline (flagged for the follow-up flag cleanup).
- Remove the `load_E_from_flowsa` field from `USAConfig` and its key from 37 config YAMLs; `derive_E_usa()` now just returns `load_E_from_flowsa()`.

Kept in place (still imported by flowsa FBA/FBS builders `EIA_MECS`, `BEA_CEDA_*`, `EPA_GHGI`): the seven MECS energy/non-energy-fuel `co2/` modules, `other_gases/common_ratios.py`, `transportation_fuel_use/`, `mappings/{cornerstone,v7}`, and `allocation/utils.py`.

Behavior change is confined to integration tests: configs with no GHG method flag (e.g. `test_usa_config_waste_disagg`) now resolve E via the `GHG_national_CEDA_2023` FBS fall-through instead of in-process allocators.

## Testing

Full unit suite green (228 passed); ruff + mypy clean. `derive_E_usa()` under `2025_usa_cornerstone_v0_3` matches the pinned `E_USA_ES` GCS snapshot exactly (max abs diff 0.0). Smoke-tested the `GHG_national_CEDA_2023` fall-through under `test_usa_config_waste_disagg` (7×405, ~4.9 Gt CO2e) and all seven FBA-builder keeper imports.


bl-young:
This might be similar to #522. We may need to consolidate

briantobin-99:
> This might be similar to #522. We may need to consolidate

Yeah there's significant overlap. Let's land your PR and then I can adjust this one to have any remaining clean up

PR #544, opened 2026-07-27
Remove load_E_from_flowsa flag and dead allocation leftovers

briantobin-99:
cc:
Closes:

## What changed? Why?

`derive_E_usa()` loads E from a flowsa FBS unconditionally since #522; the `load_E_from_flowsa` flag no longer gates anything. This removes the flag and the remaining dead files:

- Remove the `load_E_from_flowsa` field from `USAConfig` and its key from every config YAML that still set it.
- Delete `extract/allocation/{usda,utils}.py` — their only importers were the registry allocators removed in #522.
- Delete `transform/allocation/mappings/v5/` (513 lines, zero importers; the live MECS maps are `v7` + `cornerstone`).
- Remove the orphaned `E_usa_es_snapshot` fixture from `transform/conftest.py` — its only consumers were the gas allocator suites removed in #522.

Configs with no GHG method flag (e.g. `test_usa_config_waste_disagg`) resolve E via the `GHG_national_CEDA_2023` FBS fall-through, same as before — the flag was already ignored.

## Testing

Unit suite, black, ruff, and mypy green. `derive_E_usa()` under `2025_usa_cornerstone_v0_3` matches the pinned `E_USA_ES` snapshot exactly; the `GHG_national_CEDA_2023` fall-through was smoke-tested (7 gases × 405 sectors). Snapshot integration suite ([test_integration run](https://github.com/cornerstone-data/bedrock/actions/runs/30572329150)) passes on the stack including this change.


bl-young:
This makes sense to me

bl-young (approved):
Yep dropping this flag makes sense. And those usda extract functions were tied to building the flowsa version of the ceda GHG model so makes sense they can be removed as well.

PR #545, opened 2026-07-27
Purge historical configs, parity tests, and settled studies

briantobin-99:
cc:
Closes:

## What changed? Why?

Deletes the config YAMLs that only existed to re-run historical release scenarios, plus the two dispatchers that loaded them. Re-running these scenarios is retired; comparability is preserved through the frozen GCS snapshots (`snapshot_version_or_git_sha`) and the pinned Drive sheet registries, and everything remains recoverable from git history.

Deleted configs (21 of 45):

- `2025_usa_cornerstone_full_model_v0_3_*` (6) — v0.3 release stepping stones
- `useeio_phoebe_23_*` variants (7) — single-toggle USEEIO-parity configs (`useeio_phoebe_23.yaml` itself stays for the margin comparisons)
- `2025_usa_cornerstone_v0_2.yaml` — v0.2 is frozen to its snapshot (`7372464`), no longer runnable
- `2025_usa_cornerstone_{fbs_schema,ghg,B_transformation_and_waste_disaggregation,taxonomy_and_waste_disagg}` — referenced only as string merge keys in `combinations.py`, which keep working without the files
- `*_A_summary_tables_realigned` (2) — zero references

Deleted code that only served the above: the two `analysis/v0_3` dispatchers (the plotters and pinned sheet registries stay, so historical sheets can still be combined and re-plotted), `test_sef_vs_useeio_baseline.py`, and two phoebe-only waste-weight-selection tests.

The `v03_waterfall_*` configs are kept so the flag migrations in the stacked PRs can be tracked against them.

Repointed `2025_usa_cornerstone_v0_2` consumers to `2025_usa_cornerstone_v0_3`: `test_sef_phi_wiring`, two waste-disagg and one electricity regression test, `export_cornerstone_waste_disagg_matrices`, the four `baseline_snapshot_comparison` scripts, `compare_margin_approaches`, and `compare_make_use_by_stages`. A stale `__main__` block in `transform/allocation/derived.py` and the dead config menu in `scripts/build_model_manual_check.py` are cleaned up.

## Testing

Unit suite, black, ruff, and mypy green. Snapshot integration suite ([test_integration run](https://github.com/cornerstone-data/bedrock/actions/runs/30572329150)) passes on the stack including this change.3 snapshots.


bl-young:
A few initial reactions:

- `useeio` configs. I believe I am fine w/ removing these and therefore in the next PR the ability to recreate the useeio model within bedrock. We have those models and diagnostics created if we need them in the future, which I doubt we will need to do.
- the margins work - some of that was added after v0.3.0 was tagged (i.e., this is a Q3 project). In general I would recommend that we don't remove things that haven't been included in a release. If we need to go back, I think going back to a specific version is going to be easier than trying to trace back the commit history. However I don't feel strongly and recognize that keeping this is going to prevent removal of some of the gates. I'll defer to @WesIngwersen on this one.
- Same reaction to the `reconciling_data_years` project which was pulled in after v0.3.0 (though I see we have a [specific tag](https://github.com/cornerstone-data/bedrock/releases/tag/reconciling_data_years_final) for that).
- `v0.3 analysis` - I see that only the dispatchers were removed for this, which makes sense to me as we can't re-dispatch but the supporting plotting functions and such remain.

bl-young:
This PR proposes to remove the configs that we used for the v0.3 waterfalls. I am thinking ahead to our next set of diagnostics for the v0.4 release. If we are confident that all new changes (nowcasting, ABSR, disaggregation, etc.) will be stacked as new bars on top of the ones we've already displayed, then I don't think we need to keep those configs and their flags. 

HOWEVER, if there is a chance we might re-shuffle how we display the bars, I'd be more worried about removing some of these gates.

bl-young:
The specific example we discussed in the meeting today was we might want to bucket nowcasting in with methodology reconciliation (or as a new bar) before the 2024 data update for the US model out of bedrock.

bl-young:
Thanks for these changes. What would you think about keeping the `v03_waterfall` configs for a bit longer so we can track the change in flag names as they relate to these specific configs in subsequent PRs?

bl-young (approved):
Looks good thanks for the adjustment @briantobin-99

PR #550, opened 2026-07-30
plan: let the FBS assign directly to non-NAICS BEA sectors

WesIngwersen:
!! **PR to enable commenting on plan only. Do not merge** !! 

[bea_code_space_cleanup.md](../tree/plan_bea_code_space/.claude/plan/bea_code_space_cleanup.md)

Written from a day's worth of individually-diagnosed defects on nipa_fd_allocation_fix that all reduced to one cause: the pipeline treats "sector" and "NAICS code" as the same thing, so a BEA detail code that is not a NAICS code cannot be an attribution target. Its value is then dropped, or renormalised onto whichever sibling happened to be reachable.

Of the 406 BEA 2017 detail codes, 210 cannot be reached through any NAICS level - BEA detail is mostly aggregates (611A00, 813100, 5241XX, 52A000) that are not NAICS codes at all. NIPA_FD is the prime example rather than the whole scope.

The plan inventories the eight places that assume NAICS (the "NAICS guards"), notes that industry_spec_key is the only one that knows non_naics exists, and splits the work into three phases: fix the pipeline, then move NIPA_FD onto it, then retire the same workarounds in Detail_* and CAP_HAP_*.

It also records the current blocker. Removing non-NAICS BEA codes from NAICS_2017_Crosswalk.csv breaks GHG_national_Cornerstone_2024: the grand total is identical to the cent while 15 sectors vanish and 3920 of 4034 groups move. That crosswalk appears to carry the NAICS hierarchy the aggregation machinery walks, not just the target list - to be confirmed against three unexamined functions in naics.py before anything else proceeds.

No code here, and the working state it describes is uncommitted and currently GHG-breaking. Landing the plan on its own branch off main so it is readable without carrying that state.

### DIFF
--- a/.claude/plan/bea_code_space_cleanup.md
+++ b/.claude/plan/bea_code_space_cleanup.md
@@ -1,4 +1,4 @@
-# Plan: Revise FBS pipeline to work seamlessly for direct BEA assignment withou passing though NAICS guards
+# Plan: Revise FBS pipeline to work seamlessly for direct BEA assignment without passing though NAICS guards
 
 Written 2026-07-29 from the working state on `nipa_fd_allocation_fix`.
 
--- a/.claude/plan/bea_code_space_cleanup.md
+++ b/.claude/plan/bea_code_space_cleanup.md
@@ -10,8 +10,7 @@ For FBS creation that targets BEA schema sectors that are not NAICS, be able to 
 indexed by BEA 2017 detail commodity. Today the FBS is built in NAICS space and
 converted afterwards by `map_fbs_sectors_to_model_schema`. That conversion is lossy in
 both directions and is the common cause behind a long series of individually-diagnosed
-bugs. The goal is for the FBS to be natively BEA, end to end, after which that call can
-come out of `nowcast.py` entirely.
+bugs. The goal is for an FBS to able to be natively BEA, end to end.
 
 Three parts, all required
 together:
@@ -38,7 +37,6 @@ The scale of it: of the 406 BEA detail codes in
 `NAICS_Crosswalk_BEA_2017_Detail.csv`, **210 cannot be reached through any NAICS level**.
 BEA detail is mostly aggregates (`611A00`, `813100`, `5241XX`, `52A000`, `1111A0`) that are
 not NAICS codes at all.
-
 
 `d55e8e0` is the enabling fix for everything below. `assign_technological_correlation`
 left-merges the naics key against `Sector_Levels` and then does
--- a/.claude/plan/bea_code_space_cleanup.md
+++ b/.claude/plan/bea_code_space_cleanup.md
@@ -142,7 +142,7 @@ consistently downstream of `industry_spec_key`.
 | sector levels | `flowbysector.py:287` | `Sector_Levels` covers the code | unexamined |
 | crosswalk source year | `flowbyactivity.py` ~520 | `SectorSourceName` parses as `NAICS_<year>_Code`, then rows are filtered to that literal | forces every activity-to-sector crosswalk to *declare* NAICS even when its sectors are BEA codes |
 | activity schema branch | `flowbyactivity.py` ~486 | `"NAICS" in activity_schema` selects the whole mapping path | non-NAICS sources take the crosswalk path by accident, not by design |
-| model schema mapping | `allocation/derived.py:200` | NAICS_3/4/5 expand to NAICS_6 before mapping | to be removed for `nowcast.py`, see ordering |
+| model schema mapping | `allocation/derived.py:200` | NAICS_3/4/5 expand to NAICS_6 before mapping | can break non-naics if expansion required |
 | out-of-tree | `analysis/time_series_B_matrix/derive_B_time_series.py:257` | reads the NAICS crosswalk directly | check, do not assume |
 
 The `SectorSourceName` guard is the clearest illustration of the problem. The new
--- a/.claude/plan/bea_code_space_cleanup.md
+++ b/.claude/plan/bea_code_space_cleanup.md
@@ -8,8 +8,8 @@ For FBS creation that targets BEA schema sectors that are not NAICS, be able to 
 ## Prime example
 `NIPA_FD_*` builds the final-demand section of the Use table, so its output has to be
 indexed by BEA 2017 detail commodity. Today the FBS is built in NAICS space and
-converted afterwards by `map_fbs_sectors_to_model_schema`. That conversion is lossy in
-both directions and is the common cause behind a long series of individually-diagnosed
+converted afterwards by `map_fbs_sectors_to_model_schema`. That conversion is lossy when input data are originally in BEA schema 
+and is the common cause behind a long series of individually-diagnosed
 bugs. The goal is for an FBS to able to be natively BEA, end to end.
 
 Three parts, all required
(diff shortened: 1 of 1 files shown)


.claude/plan/bea_code_space_cleanup.md
```diff
@@ -0,0 +1,257 @@
...
+
+### The blocker: this breaks GHG
+
+`GHG_national_Cornerstone_2024` before vs after: totals identical to the cent, but 8695 →
+8355 rows, 3920 of 4034 groups moved, 466 sectors affected, and **15 sectors vanish**:
+
+```
+230301 230302 233210 233230 233240 233262 2332A0 2332C0 2332D0 233411 233412 2334A0
+531HSO 531HST 531ORE
+```
```
bl-young, line 85: Yes these are the non-naics codes that were added to the Cornerstone FBS in order to explicitly avoid the BEA -> NAICS -> BEA issue

.claude/plan/bea_code_space_cleanup.md
```diff
@@ -0,0 +1,257 @@
+# Plan: Revise FBS pipeline to work seamlessly for direct BEA assignment withou passing though NAICS guards
+
+Written 2026-07-29 from the working state on `nipa_fd_allocation_fix`.
+
+## The goal
+For FBS creation that targets BEA schema sectors that are not NAICS, be able to assign/attribute directly to them and avoid translating the code to NAICS. Clean up existing code that forces NAICS on non-NAICS based codes.
+
```
bl-young, line 7: I agree with this goal for select sectors - and we tried to introduce the workaround for the GHG model to enable this. I expect it was not fully comprehensive for what you are aiming to do w/ the NIPA model

.claude/plan/bea_code_space_cleanup.md
```diff
@@ -0,0 +1,257 @@
...
+## Prime example
+`NIPA_FD_*` builds the final-demand section of the Use table, so its output has to be
+indexed by BEA 2017 detail commodity. Today the FBS is built in NAICS space and
+converted afterwards by `map_fbs_sectors_to_model_schema`. That conversion is lossy in
+both directions and is the common cause behind a long series of individually-diagnosed
+bugs. The goal is for the FBS to be natively BEA, end to end, after which that call can
```
bl-young: To be clear, do we think this is especially problematic for the NIPA FBS particularly, because the NIPA codes have such unique mappings across BEA-like codes? I guess I'm generally recalling that outside of a few specific areas, the BEA -> NAICS mapping is generally pretty clean so I would not typically say its "lossy in both directions"

bl-young:
My general comment is that 
1) yes I expect the handling of non-naics sectors was not implemented in a comprehensive way - a bit hacky for the GHG model, so I am not entirely surprised it is giving your problems.
2) it's not clear to me if you are suggesting that the GHG model as it currently stands on main is being calculated incorrectly, or if that the fix you are proposing implementing means we will have to adjust the GHG model to get it working correctly again (or perhaps still not sure and we need to look at these non-naics sectors like construction a bit closer).

.claude/plan/bea_code_space_cleanup.md
```diff
@@ -0,0 +1,257 @@
...
+   shifts sectors *into* the Cornerstone schema — 12 of its 409 activities are not BEA
+   detail codes (`562HAZ`, `562OTH`, `562111`, `562212`, `562213`, `562910`, `562920`,
+   `331314`, `S00101`, `S00201`, `S00202`, `F01000`), Cornerstone's waste disaggregation
+   being why the whole 562 family has been pathological.
+8. Roll out to `NIPA_FD_2018`–`2024`, ideally via the `NIPA_FD_common.yaml` refactor
+   (#503) rather than eight copies.
+
+### Phase 3 — retire the workarounds elsewhere
+
+9. `Detail_Use_SUT`, `Detail_Supply`, `Detail_Make`, `Detail_Use` and the 13 `CAP_HAP_*`
```
bl-young, line 197: It's been a while since we've touched the `CAP_HAP` method, and the `Detail_*` FBS are not used anymore for the GHG model.

All of which is to say, we should definitely make a note in those methods about this needed change, or flag that those sectors will need to be adjusted if they are used in the future.

bl-young:
> 2. it's not clear to me if you are suggesting that the GHG model as it currently stands on main is being calculated incorrectly, or if that the fix you are proposing implementing means we will have to adjust the GHG model to get it working correctly again (or perhaps still not sure and we need to look at these non-naics sectors like construction a bit closer).

Based on what i'm seeng here:
> 4. **Gate:** GHG A/B returns to zero delta *with the BEA codes present*, and the `'531'`/`'23'`/`'92'` entries can come out of the `NAICS_6` lists without anything moving. Nothing in Phase 2 lands before this passes.

I think the answer to this is that we expect, once fixed, the GHG model results will stay as is.

.claude/plan/bea_code_space_cleanup.md
WesIngwersen: Yes but they are all under `NAICS_6`. This leads to them having to be added in crosswalks as NAICS sectors, even when the 531* are not.  That caused problems for data coming in associated with the BEA codes.

https://github.com/cornerstone-data/bedrock/blob/42f7e596e34e3284f42eb5ebef7a4362aa79237d/bedrock/transform/common/Cornerstone_2025_target.yaml#L28-L54

WesIngwersen:
> My general comment is that
> 
> 1. yes I expect the handling of non-naics sectors was not implemented in a comprehensive way - a bit hacky for the GHG model, so I am not entirely surprised it is giving your problems.
> 2. it's not clear to me if you are suggesting that the GHG model as it currently stands on main is being calculated incorrectly, or if that the fix you are proposing implementing means we will have to adjust the GHG model to get it working correctly again (or perhaps still not sure and we need to look at these non-naics sectors like construction a bit closer).

2. I am not suggesting the GHG model has errors now. But because it implements the same BEA as NAICS hacks, this refactor would involved editing the method.  Hopefully the outcome is still the same.

.claude/plan/bea_code_space_cleanup.md
WesIngwersen: It would be cleanest to go ahead and edit any active method to make sure it works with the new approach.  As part of the broader cleanup way beyond this project I would say we should remove all non-used non-tested FBA and FBS methods and keep them archived in flowsa to bring in as needed.

WesIngwersen:
Just one comment to add.. I think this would be a good direction more broadly to go with flowsa-inherited code because we'd like to expand it to be able to process data from other regions.  This project would not make the code schema independent by a long-shot, but its moving it in that direction.

catherinebirney:
I think it is a good idea to make the code more generic/flexible and less NAICS dependent. I like the idea in the plan of not requiring the `non_naics` to have parent/child relationships and instead be standalone. We have unnecessary sector assignments to get the household and gov codes to be "6-digit" or "2-digit" NAICS when they are both a single 6 length string. 

I remember we looked into allowing a full BEA schema a few years ago and it does require a pretty big overhaul of the code. Which is also what this plan is capturing.

WesIngwersen:
@catherinebirney @bl-young Thanks for your feedback. OK I will plan to start moving forward on this. > will close this PR and likely rename this branch to work from and keep a plan in that branch.

PR #553, opened 2026-07-30
Introduce use_cornerstone_ghg_model bucket flag (GHG model allocation)

briantobin-99:
cc:
Closes:

## What changed? Why?

Introduces the "GHG model allocation" bucket flag and dissolves the year-locked inventory flags into the data-year knob:

- New `use_cornerstone_ghg_model`: ON = load the pre-built `GHG_national_Cornerstone_{usa_ghg_data_year}` FBS parquet from GCS (with the eGRID variant when `implement_electricity_disaggregation` is on); OFF = the `GHG_national_CEDA_{year}` flowsa FBS (legacy CEDA allocation methodology; raises a clear error for years ≠ 2023, where no CEDA method file exists).
- Absorbs `new_ghg_method` (identical mechanism) and `v0_3_umd_2024_ghgia` (the same mechanism year-locked to 2024). Which inventory and attribution vintages a year carries (EPA GHGI vs UMD GHGIA, MECS survey year) is defined by the per-year method files in `transform/ghg/`, not by config flags.
- Deletes `update_mecs_method` and `v0_3_umd_2023_ghgia` — intermediate 2023 inventory states that never shipped — along with their now-unselectable variant method files (`GHG_national_Cornerstone_2023_{ghgi_mecs,umd_ghgia}.yaml`) and the duplicated year-guard blocks (`_select_flowsa_ghg_method` is gone; selection lives in one place in `load_E_from_flowsa`).
- The mutual-exclusion validator with `use_ghg_national_2023_m2` carries over to the new flag.

Config migration: `2025_usa_cornerstone_v0_3.yaml`, the `v03_waterfall_*` configs, and the electricity configs swap their `new_ghg_method` / `v0_3_umd_2024_ghgia` keys for `use_cornerstone_ghg_model`. Under the new scheme, the waterfall G2→G3 diff ("US data update") is purely `usa_ghg_data_year: 2023 → 2024`.

## Testing

Unit suite, black, ruff, and mypy green. `derive_E_usa()` under `2025_usa_cornerstone_v0_3` matches the pinned `E_USA_ES` snapshot exactly (max abs diff 0.0). Snapshot integration suite ([test_integration run](https://github.com/cornerstone-data/bedrock/actions/runs/30572329150)) passes on the stack including this change.


bedrock/utils/config/usa_config.py
```diff
@@ -91,10 +91,9 @@ class USAConfig(BaseModel):
     useeio_margins: bool = False  # DRI: WesIngwersen
     cornerstone_industry_avg_margins: bool = False  # DRI: WesIngwersen
     ### GHG Methodology selection
-    new_ghg_method: bool = False  # if True, it is the new Cornerstone GHG FBS
-    update_mecs_method: bool = False  # DRI: catherine.birney
-    v0_3_umd_2023_ghgia: bool = False  # DRI: catherine.birney
-    v0_3_umd_2024_ghgia: bool = False  # DRI: catherine.birney
+    # "GHG model allocation" bucket: Cornerstone GHG FBS (pre-built parquet at
+    # usa_ghg_data_year) vs the legacy CEDA-methodology FBS (2023 only).
+    use_cornerstone_ghg_model: bool = False
```
bl-young, line 96: Should this be true by default?

bedrock/utils/config/configs/useeio_phoebe_23.yaml
```diff
@@ -28,7 +28,6 @@ deflate_x_to_detail_io_year_for_B: true
 use_ghg_national_2023_m2: true
 
 # Must remain false whenever use_ghg_national_2023_m2 is true.
-new_ghg_method: false
```
bl-young, line 31: if we do set the default to be True this will need to be revised

bl-young:
Outside of those comments/questions above, this looks good to me

bedrock/utils/config/usa_config.py
bl-young: or maybe we will make these switches later. I don't feel strongly.

bl-young (approved):

jvendries:
Looks good to me as well.

PR #555, opened 2026-07-30
Delete USEEIO-recreation flags without ongoing consumers

briantobin-99:
cc:
Closes:

## What changed? Why?

Deletes the USEEIO-recreation flags that have no remaining consumer, plus their gated code:

- `use_ghg_national_2023_m2` — the m2 FBS path, the output-weighted NAICS→BEA mapping, and the CH4-GWP-27.9 workbook override. The electricity NAICS retargeting hook in the standard mapping path is untouched.
- `useeio_margins` — the USEEIOR margin filter set and `inflation_helpers_useeio.py` (the USEEIOR CPI table). `get_rho_inflation_ratio` stays (the publish Excel Rho panel uses it).
- `ceda_margins` — the legacy CEDA filter set and `derive_2017_margins_ceda_usa` / `derive_phi_ceda_usa`.
- `use_useeio_schema` — read by no code; its only effect was the m2 validator interlock.

Kept: `scale_a_matrix_with_useeio_method`, `use_E_data_year_for_x_in_B`, and `deflate_x_to_detail_io_year_for_B` remain in use as diagnostic probes by the electricity-disaggregation work (`year_alignment_bly_e.py`, the electricity configs, and the mixed-units handling inside the `D/N_new_inflated` block), so they are out of scope here.

Configs deleted: `useeio_phoebe_23.yaml` and `v03_waterfall_useeio_g1_schema_ghg.yaml` — both set the deleted margins/m2 flags; the USEEIO-track waterfall sheets stay pinned in the registry, and the CEDA-track waterfall configs are untouched.

Margins comparison scripts: `compare_margin_approaches` and `compare_phi_to_reference` (studies of the settled margins decision) are deleted, with an archived output run stored externally as the record. The two `compare_sef_*` scripts are kept on pinned inputs: `--phoebe-sef-csv` is now required (a phoebe SEF published before this merges serves as the pin), and `compare_sef_zenodo_useeio_code` defaults to `2025_usa_cornerstone_v0_3`.

## Testing

Unit suite, black, ruff, and mypy green. None of the deleted paths are active in the v0.3 config; Snapshot integration suite ([test_integration run](https://github.com/cornerstone-data/bedrock/actions/runs/30572329150)) passes on the stack including this change.


WesIngwersen:
> **Margins-project handling (the sign-off question):** `compare_margin_approaches` and `compare_phi_to_reference` (June studies of the settled margins decision) are deleted. the two recent `compare_sef_*` scripts are kept and converted to pinned inputs: `--phoebe-sef-csv` becomes required (publish a phoebe SEF once before merge and pin it), and `compare_sef_zenodo_useeio_code` defaults to `2025_usa_cornerstone_v0_3`. If live phoebe publishing is still needed for Q3 margins work, hold this PR.

I'm making sure we have a record of the margins analysis output linked to a git commit which we can store internally so we have the analysis...i will open up a PR with updates i made to be able to run it off main; store the outputs on GDrive, then I am fine with deleting the margins analysis code and configs stated in this PR, as long as the team is happy with the record of output being enough data/visuals to explain margin changes from useeio and ceda.

WesIngwersen:
FYI @briantobin-99 please review and pull this one in before merging this.  Otherwise I am good with this one.

bedrock/utils/config/configs/v03_waterfall_useeio_g1_schema_ghg.yaml
bl-young, line 1: I'm questioning the decision to delete one of our waterfall configs. But perhaps I'll look at 559 next to ensure this is ok

bl-young:
Outside my comment above this looks good.

bl-young (approved):

jvendries:
Minor comment - PR description still says margins/G1 are deleted but that is outdated after the restore commit.
Otherwise looks good.

PR #556, opened 2026-07-30
fix margin analysis; update README to run in current env

WesIngwersen:
merge before PR #555 

## What changed? Why?
Margins analysis needed to be updated to run in current environment and README updated to clearly tie a script to an output before archiving. 

## Testing

Reran analysis scripts and verified output.
Output stored on CS drive https://drive.google.com/drive/folders/1z984rj3hM-4dxJPYpqAxP7na48nRj8Wd


briantobin-99 (approved):
LGTM

PR #559, opened 2026-07-30
Add waterfall-progression regression test (incremental bucket flags)

briantobin-99:
cc:
Closes:

## What changed? Why?

Adds an integration test guaranteeing the release-waterfall progression keeps working, and fixes a B year-scaling regression on the legacy-footing configs.

**Test.** Toggling the bucket flags incrementally (v0 baseline → `use_cornerstone_ghg_model` → `+ implement_waste_disaggregation` → `+ apply_io_year_adjustments` + margins → `usa_ghg_data_year: 2024`, via the `v03_waterfall_*` configs) must reproduce each step's q-weighted average total emission factor — the release waterfall convention:

wavg N = Σᵢ Nᵢ qᵢ / Σᵢ qᵢ, with N = 1ᵀ B L (total-requirements L) and each step weighted by its own run's gross-output vector q.

- New `bedrock/utils/validation/waterfall_progression.py` — computes wavg N for one config, deriving N exactly as the EF diagnostics derive `N_new`, with a CLI and a `--v0-baseline` mode recomputing the baseline from the frozen v0 snapshots.
- New `test_waterfall_progression.py` (`eeio_integration`-marked, so it runs on the twice-weekday scheduled integration workflow) — runs each step in a fresh subprocess so `functools.cache` state cannot leak between configs, and asserts each step's level against pinned expected values whose per-sector N vectors match the `N_and_diffs` tabs of the pinned diagnostics sheets in `release_v0_v03_ceda_groups` (the sheets behind the v0.3 assessment) to ~5e-10 across all 405 sectors. Tolerance is 1e-4 kgCO2e/USD against step bars of ~1e-2, and `v03_waterfall_final` must telescope to the last step. The USEEIO-baseline chain is covered too: it shares the G2/G3 states with the CEDA chain, and its one extra state (`v03_waterfall_useeio_g1_schema_ghg`) is pinned as a seventh step, so both release-waterfall chains stay reproducible.

**Fix.** `derive_cornerstone_B_non_finetuned` keys the legacy-footing B year treatment on the bucket vocabulary: when `use_ghg_year_x_in_B` is false, B is scaled to `usa_io_data_year` (`scale_cornerstone_B`) and inflated to `model_base_year` (`inflate_cornerstone_B_matrix_with_industry_pi`); when true, B stays plain-vnorm. This restores the pre-#482 semantics for legacy-footing configs, which #482 had inadvertently dropped by re-keying the treatment onto the experimental `use_scaled_x_and_scaled_Vnorm_for_B` flag. The v0.3 path is unaffected.

## Testing

All six waterfall tests pass, and running with `USA_CONFIG_FILE` exported into the environment (the scheduled-workflow condition, where earlier suite tests in the same process set it) produces identical values. Unit suite, black, ruff, and mypy green.

### DIFF
--- a/bedrock/transform/__tests__/test_waterfall_progression.py
+++ b/bedrock/transform/__tests__/test_waterfall_progression.py
@@ -1,0 +1,97 @@
+"""Waterfall-progression regression: incremental bucket flags reproduce the release steps.
+
+The v0→v0.3 release waterfall's bedrock-side steps are, in bucket-flag
+vocabulary (the ``v03_waterfall_*`` configs):
+
+    v0 snapshot baseline
+      → + use_cornerstone_ghg_model            (G1a: "GHG model allocation")
+      → + implement_waste_disaggregation       (G1b: "Waste disagg.")
+      → + apply_io_year_adjustments + margins  (G2:  "IO year adjustments")
+      → usa_ghg_data_year: 2023 → 2024         (G3:  "US data update")
+
+Each test derives one step's total attributed emissions ΣBLy = Σ diag(d)·L·y
+in a fresh subprocess (so ``functools.cache`` state cannot leak between
+configs) and compares against the totals of the ``BLy_new_vs_BLy_old`` tabs
+of the pinned diagnostics sheets in
+``bedrock.utils.validation.analysis.release_v0_v03_ceda_groups`` — the same
+sheets the v0.3 assessment waterfall figure was built from.
+
+Note on the published figure: its bars (baseline 5,069; −42/+16/−254/+22)
+are the USA *portion of the global MRIO* after ingesting these bedrock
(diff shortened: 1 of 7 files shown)


bl-young:
I like this idea (sorry I know its still in draft), but would it make sense to build this test around the weighted average N waterfalls instead of the BLy waterfalls that we shifted away from?

bl-young:
Do we also need to test/maintain the useeio progression as well for this test and not just the ceda/v0 baseline? Which I guess gets back to my comment on #555 of the config that got deleted

bl-young:
Given the change in the waterfall approach, I think this needs to be stacked on top of #555 instead of its current spot as being adjacent to 555.

bl-young:
This has been replaced by #594

PR #561, opened 2026-07-31
refactor(config): hardcode the Cornerstone model schema, delete use_cornerstone_2026_model_schema

briantobin-99:
cc:
Closes:

## What changed? Why?

`use_cornerstone_2026_model_schema` is deleted. Every config set it True — the Cornerstone 2026 model schema is the only supported schema — so all code paths now assume the flag's True side unconditionally. 40 files, net −719 lines.

**Model pipeline.**
- `transform/eeio/derived.py`: the legacy wrappers collapse to their cornerstone implementations; `derive_B_usa_via_vnorm` and `scale_abq_via_summary.py` (orphaned legacy-B helpers) are deleted, along with the unused `inflate_A_matrix`/`inflate_B_matrix`/`inflate_q_or_y` helpers in `inflation_helpers_ceda.py`.
- `transform/allocation/derived.py`: `map_fbs_sectors_to_model_schema` always maps into `Cornerstone_2025` (the electricity-disaggregation retargeting hook is unchanged); the E collapse reindexes to the Cornerstone columns directly, and the legacy BEA→CEDA groupby collapse is gone.

**Allocator chain.** The MECS mapping selectors, `get_allocation_sectors()`, and `load_bea_use_table()` are pinned to the Cornerstone frame — the side that generated the pinned `*_CEDA` allocation FBA parquets on GCS, which carry Cornerstone-frame activity codes (`33131B`, `335220`, disaggregated `562*`). The use-table alignment helpers for CEDA-vocabulary sectors are unchanged.

**Diagnostics.** EF diagnostics always align old (CEDA v7) vs new (Cornerstone) EF vectors, always emit `comparison_type` and the `sector_mapping_notes` tab.

**Guards and shims.** `require_cornerstone_config` (publish) and the `_force_cornerstone_model_schema` context manager (nowcast) are deleted with their call sites — they only asserted or forced the flag.

**Config.** The field is removed from `USAConfig` and the key stripped from all config YAMLs. Behavior-neutral: every YAML set it True, and `USAConfig`'s strict validation requires removing the key in the same change.

## Testing

- Unit suite, black, ruff, and mypy green.
- The v0.3 snapshot integration tests (`test_usa.py`, 9 tests) pass against the pinned snapshots.
- The allocation FBA builders were re-run end-to-end and reproduce the pinned GCS parquets bit-identically: `BEA_CEDA_foam_allocation_2023`, `BEA_CEDA_refrigeration_allocation_2023`, and `EIA_MECS_Energy_Allocation_CEDA_2018` (2,835 rows; rebuilt under `usa_ghg_data_year: 2023`, the vintage the pin was generated with — the EPA allocation tables used by the MECS builders do not yet support 2024).

Snapshot integration runs for this branch: https://github.com/cornerstone-data/bedrock/actions/runs/30674351415 (green) and https://github.com/cornerstone-data/bedrock/actions/runs/30948864495 (after adding the USEEIO-chain waterfall step).


bl-young:
I generally agree with the direction this is taking. I'd like to reserve further judgement until my questions on 555 and 559 are resolved.

bedrock/transform/eeio/derived.py
```diff
@@ -75,267 +40,38 @@
...
-            target_year=get_usa_config().model_base_year,
-        )
-
-        return SingleRegionYtotAndTradeVectorSet(
-            ytot=YVectorSchema.validate(ytot),
-            exports=ExportsVectorSchema.validate(exports),
-            imports=ImportsVectorSchema.validate(imports),
-        )
+    """Detail Y and trade vectors, structurally reflected to the target year."""
+    return derive_cornerstone_Y_and_trade_scaled()
```
bl-young, line 57: these kind of pass through functions may be due for refactoring/renaming in the future.

bedrock/transform/eeio/derived.py
```diff
@@ -75,267 +40,38 @@
...
-
-        # NOTE: original y values have some negative values
-        # that will distort the scaling process that uses the ytot here in non-US countries.
-        # We make a data assumption here to set them to 0 in order to make the scaling process valid.
-        # TODO: this is a temporary solution, we need a y pandera type to enforce this data assumption.
-        return YVectorSchema.validate(
-            handle_negative_vector_values(y_national_acct_balance_detail_2022)
-        )
+    """y for the national accounting balance: y_nab = (y_tot - y_imp) + exports."""
+    return derive_cornerstone_y_nab()
```
bl-young, line 63: this one as well

bl-young (approved):

Issue #436, opened 2026-05-26
Make vs Use output totals do not match for specific industry and commodity sectors in the Cornerstone pipeline


jvendries:
### Summary

EEIO validation checks that gross output implied by the **Make** table (`V`) matches output implied by the **Use** table (`U`), value added (`VA`), and trade/final demand (`Ytot`). The **Cornerstone** pipelines fails this check for a small set of sectors at a 5% relative tolerance. Failures are **not** due to CPI scaling (CPI consistency tests pass) and appear to stem from **inconsistent treatment of Make vs Use paths** during IoT redefinition, BEA→Cornerstone correspondence, and waste disaggregation.

These failures are currently marked as **expected xfails** in integration tests. This issue documents the mismatch for method reconciliation.

<details> 
<summary> Analysis </summary> 

### What is being checked

Validation logic lives in `compare_output_from_make_and_use` (`bedrock/utils/validation/eeio_diagnostics.py`):

| Check | Make side | Use + other side |
|-------|-----------|------------------|
| **Industry output** (`x`) | `V.sum(axis=1)` — row sums of Make | `U.sum(axis=0) + VA.sum(axis=0)` — column sums of Use plus VA |
| **Commodity output** (`q`) | `V.sum(axis=0)` — column sums of Make | `U.sum(axis=1) + (ytot + exports − imports)` — row sums of Use plus net final demand from trade |

Both sides should agree within tolerance if Make and Use are built consistently from the same underlying BEA detail and transformations.

#### Cornerstone (`test_eeio_accounting.py`, xfails)

**Commodity (4 failing sectors)**
- Sectors: `562HAZ`, `562212`, `562OTH`, `S00402`
- **Hypothesis:** Waste disaggregation weights are calculated based on an imbalanced table. Test for this is to re-calculate the weights once the after-correspondence tables have been balanced.

**Industry (10 failing sectors)**
- Includes waste codes above plus `331110`, `331314`, `331510`, `332119`, `481000`, `523A00`, `812200`
- **Hypothesis:** BEA→Cornerstone correspondence drops the values for the Scrap (`S00401`), Noncomparable imports (`S00300`), and RoW adjustment (`S00900`) commodities for both the Make and Use tables by zeroing them out (columns and rows, respectively). However, the values removed are _not the same_ for both tables:

| Commodity| Make column total dropped| Commodity row total dropped |
|-------|----------|-----------|
| Scrap (`S00401`)| $10.8B| $24.2B |
[...]

PR #562, opened 2026-08-01
Add compare_E_and_LCI_result test analogous to the same test in useeior

jvendries:
cc:
Closes:

## What changed? Why?

Adds `compare_E_and_LCI_result` — a port of useeior's `compareEandLCIResult` — which validates that the direct-perspective LCI matrix (`B·diag(L·y)`) matches the commodity-transformed satellite totals (`E_c`). Two commodity transformation paths are supported: a `Vnorm` path (used by Cornerstone's scrap-corrected market shares) and the standard `C_m` path derived from the Make table.

Introduces `eeio_year_alignment_precondition_ok` and `assert_eeio_year_alignment_precondition` to enforce that model year, GHG data year, and B-matrix construction flags are aligned before running LCI≈E checks (replacing useeior's `generateChiMatrix` with an explicit χ=1 guard). Misaligned configs raise loudly rather than silently skipping.

All new functions are exported from the `validation` package's `__all__`.

## Testing

- Unit tests cover `eeio_year_alignment_precondition_ok` passing and failing cases, `assert_eeio_year_alignment_precondition` raising on a misaligned config, and `compare_E_and_LCI_result` passing with a minimal toy LCI system and failing when `E_ind` is perturbed.
- An integration test (`test_v0_3_domestic_lci_equals_e`) runs the full domestic LCI≈E check against the live v0.3 Cornerstone pipeline using the `Vnorm` path, asserting zero failing sectors at 1% tolerance.
- A second integration test (`test_v0_3_total_lci_equals_e`) is marked `xfail` with a documented reason: the total LCI≈E check uses `y_trade` rather than IO-balanced throughput, producing the same ~1749-cell imbalance seen in total `Ly≈q`.

### DIFF
--- a/bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
+++ b/bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
@@ -515,10 +515,11 @@ def test_eeio_year_alignment_precondition_ok() -> None:
         {
             'model_base_year': 2024,
             'usa_ghg_data_year': 2024,
-            'use_E_data_year_for_x_in_B': True,
+            'apply_io_year_adjustments': True,
         },
         strict=True,
     )
+    assert cfg.use_ghg_year_x_in_B
     assert eeio_year_alignment_precondition_ok(cfg)
     assert_eeio_year_alignment_precondition(cfg)
 
@@ -528,7 +529,7 @@ def test_eeio_year_alignment_precondition_fails_loud() -> None:
         {
             'model_base_year': 2024,
             'usa_ghg_data_year': 2023,
-            'use_E_data_year_for_x_in_B': True,
+            'apply_io_year_adjustments': True,
         },
         strict=True,
     )
--- a/bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
+++ b/bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
@@ -58,14 +58,14 @@ class TestDiagnosticResult:
     def test_basic_passing_result(self) -> None:
         """Test basic instantiation with a passing result."""
         result = DiagnosticResult(
-            name='Row sum check',
+            name="Row sum check",
             passed=True,
             tolerance=0.01,
             max_rel_diff=0.005,
             failing_sectors=[],
         )
 
-        assert result.name == 'Row sum check'
+        assert result.name == "Row sum check"
         assert result.passed is True
         assert result.tolerance == 0.01
         assert result.max_rel_diff == 0.005
@@ -75,53 +75,53 @@ class TestDiagnosticResult:
     def test_failed_result_with_failing_sectors(self) -> None:
         """Test a failed result with sectors that failed the check."""
         result = DiagnosticResult(
-            name='Column sum check',
+            name="Column sum check",
             passed=False,
             tolerance=0.01,
             max_rel_diff=0.05,
-            failing_sectors=['11', '21', '31'],
+            failing_sectors=["11", "21", "31"],
         )
 
         assert result.passed is False
         assert len(result.failing_sectors) == 3
-        assert '11' in result.failing_sectors
-        assert '21' in result.failing_sectors
-        assert '31' in result.failing_sectors
+        assert "11" in result.failing_sectors
+        assert "21" in result.failing_sectors
+        assert "31" in result.failing_sectors
         assert result.max_rel_diff == 0.05
 
     def test_result_with_details_dataframe(self) -> None:
         """Test a result with a details DataFrame."""
         details_df = pd.DataFrame(
             {
-                'sector': ['11', '21'],
-                'expected': [100.0, 200.0],
-                'actual': [105.0, 195.0],
-                'rel_diff': [0.05, 0.025],
+                "sector": ["11", "21"],
+                "expected": [100.0, 200.0],
+                "actual": [105.0, 195.0],
+                "rel_diff": [0.05, 0.025],
             }
         )
 
         result = DiagnosticResult(
-            name='Detailed check',
+            name="Detailed check",
             passed=False,
             tolerance=0.01,
             max_rel_diff=0.05,
-            failing_sectors=['11'],
+            failing_sectors=["11"],
             details=details_df,
         )
 
         assert result.details is not None
         assert isinstance(result.details, pd.DataFrame)
         assert len(result.details) == 2
-        assert 'sector' in result.details.columns
-        assert 'expected' in result.details.columns
-        assert 'actual' in result.details.columns
-        assert 'rel_diff' in result.details.columns
+        assert "sector" in result.details.columns
+        assert "expected" in result.details.columns
+        assert "actual" in result.details.columns
+        assert "rel_diff" in result.details.columns
 
     def test_negative_tolerance_raises_error(self) -> None:
         """Test that negative tolerance raises ValueError."""
-        with pytest.raises(ValueError, match='Tolerance must be non-negative'):
+        with pytest.raises(ValueError, match="Tolerance must be non-negative"):
             DiagnosticResult(
-                name='Invalid check',
+                name="Invalid check",
                 passed=True,
                 tolerance=-0.01,
                 max_rel_diff=0.005,
@@ -130,9 +130,9 @@ class TestDiagnosticResult:
 
     def test_negative_max_rel_diff_raises_error(self) -> None:
         """Test that negative max_rel_diff raises ValueError."""
-        with pytest.raises(ValueError, match='max_rel_diff must be non-negative'):
+        with pytest.raises(ValueError, match="max_rel_diff must be non-negative"):
             DiagnosticResult(
-                name='Invalid check',
+                name="Invalid check",
                 passed=True,
                 tolerance=0.01,
                 max_rel_diff=-0.005,
@@ -142,7 +142,7 @@ class TestDiagnosticResult:
     def test_zero_tolerance_is_valid(self) -> None:
         """Test that zero tolerance is accepted (edge case)."""
         result = DiagnosticResult(
-            name='Exact match check',
+            name="Exact match check",
             passed=True,
             tolerance=0.0,
             max_rel_diff=0.0,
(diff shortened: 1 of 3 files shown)


jvendries:
* **#563** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/563?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#562** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/562?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/562?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jvendries:
Below is a summary of the useeior-style validation checks as they stand with the v0.3 config. 

**Summary**: There are two main failure points. 
1) `Ly ≈ q (total)`  and `LCI ≈ E (total)` fail because the `y` value used, `y_trade`, comes from the Y/trade pipeline, not the IO residual `q − A_tot q`, so it is not balanced to scaled `A_tot/q`.
2) `q ≈ U_d + y_d` and `Make vs Use` fail because of A) The BEA to Cornerstone remap and B) waste disaggregation bug that transposes that Use table intersection weights.

**Next steps**
1) Either A) validate total `Ly` against `y_tot_bal = q − A_tot q` (parity with `y_nab`), or B) rebuild/scale `ytot/trade` so empirical `y_trade` satisfies `q = A_tot q + y_trade`, then keep the trade-based check. A) is a small validation change; B) is a methodology change to the `Y/trade` pipeline.
2) A) Remap BEA to Cornerstone and/or do not drop sectors that influence commodity totals between tables; B) Fix waste disagg bug. A) is a remap of the mapping that may impact other steps in the model build; scope TBD. B) is a limited-scope fix in the waste disaggregation step.

| Validation | Function / parameters | Identity | Status on v0.3 | Explanation (failures) |
|---|---|---|---|---|
| Ly ≈ q (domestic) | `compare_output_vs_leontief_x_demand(output=q, L=L_dom, y=y_nab)` with `y_nab = derive_cornerstone_y_nab()`, `L_dom = compute_L_matrix(A=Adom)` | `L_dom @ y_nab ≈ q` | **Pass** | — |
| Ly ≈ q (total) | `compare_output_vs_leontief_x_demand(output=q, L=L_tot, y=y_trade)` with `y_trade = ytot + exports − imports` from `derive_cornerstone_Y_and_trade_scaled()`, `L_tot = compute_L_matrix(A=Adom+Aimp)` | `L_tot @ y_trade ≈ q` | **Fail** (xfailed) | `y_trade` comes from the Y/trade pipeline, not the IO residual `q − A_tot q`, so it is not balanced to scaled `A_tot`/`q` (~350 sectors). |
| LCI ≈ E (domestic) | `compare_E_and_LCI_result(B, L=L_dom, y=y_nab, E_ind, x, Vnorm, q)` (Vnorm path; χ=1 via precondition) | `B · diag(L_dom @ y_nab) ≈ E_c` | **Pass** | — |
| LCI ≈ E (total) | `compare_E_and_LCI_result(B, L=L_tot, y=y_trade, E_ind, x, Vnorm, q)` (Vnorm path; same L/y pairing as total Ly) | `B · diag(L_tot @ y_trade) ≈ E_c` | **Fail** (1749 cells; xfailed) | Same root cause as total Ly: unbalanced `y_trade` makes `L_tot @ y_trade ≉ q`, so reconstructed LCI diverges from `E_c`. |
| CPI market share | `commodity_industry_output_cpi_consistency(V, q, x, industry_CPI_ratio, commodity_CPI_ratio)` | `q · CPI_c ≈ C_m (x · CPI_i)` | **Pass** | — |
| q ≈ U_d + y_d | `compare_commodity_output_to_domestics_use_plus_exports(q, U_d, y_d)` | commodity output ≈ domestic Use + final demand | **Fail** (13 sectors; xfailed) | BEA→Cornerstone remap and waste disaggregation split parent GO across children without preserving the NAB identity sector-by-sector (mining/petroleum, 562*, S00402). |
| Make vs Use | `compare_output_from_make_and_use(output, V, U, VA, y_set)` | Make x/q vs Use+VA/FD | **Fail** (xfailed; #436) | Same remap/disagg mismatch: Make q≠Use q on 4 waste/special codes; Make x≠Use x+VA on ~10 industries. Not in default useeior suite. |

bedrock/utils/validation/eeio_diagnostics.py
```diff
@@ -465,3 +467,202 @@ def compare_output_from_make_and_use(
             'invalid output parameter requested for comparison between make and use, select commodity or industry'
         )
     return d_result
+
+
+def eeio_year_alignment_precondition_ok(cfg: USAConfig) -> bool:
```
bl-young: note these are likely to be impacted by the cleanup in the other PRs. We may want to pull those in first and then reassess

bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
```diff
@@ -454,3 +467,194 @@ def test_compare_output_and_L_y(
...
     )
     assert len(r_output_L_y_validation.failing_sectors) == 0
+
+
+# ---------------------------------------------------------------------------
+# useeior-parity LCI ≈ E + year-alignment precondition
+# ---------------------------------------------------------------------------
+
+
+@dc.dataclass(frozen=True)
```
bl-young, line 481: I'm not sure I understand the need for all these unit tests?

bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
```diff
@@ -454,3 +467,194 @@ def test_compare_output_and_L_y(
...
+    assert len(result.failing_sectors) > 0
+
+
+@pytest.mark.eeio_integration
+def test_v0_3_domestic_lci_equals_e() -> None:
+    """Live domestic LCI≈E under aligned v0.3 (Vnorm path, χ=1 precondition)."""
+    reset_usa_config(should_reset_env_var=True)
+    set_global_usa_config('2025_usa_cornerstone_v0_3.yaml')
+    try:
+        from bedrock.transform.allocation.derived import derive_E_usa
```
bl-young: why are these all inside the try block?

bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
```diff
@@ -454,3 +467,194 @@ def test_compare_output_and_L_y(
...
+    )
+    assert not result.passed
+    assert len(result.failing_sectors) > 0
+
+
+@pytest.mark.eeio_integration
+def test_v0_3_domestic_lci_equals_e() -> None:
+    """Live domestic LCI≈E under aligned v0.3 (Vnorm path, χ=1 precondition)."""
+    reset_usa_config(should_reset_env_var=True)
+    set_global_usa_config('2025_usa_cornerstone_v0_3.yaml')
```
bl-young: how are we handling the assignment of the config in other tests? Do we not just point these as integration tests to our current snapshot? Along those lines, i'm not sure we want to name this function after a certain model version.

bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
```diff
@@ -454,3 +467,194 @@ def test_compare_output_and_L_y(
...
+
+
+@pytest.mark.eeio_integration
+@pytest.mark.xfail(
+    reason=(
+        'Cornerstone total LCI≈E uses y_trade (ytot/trade), not IO-balanced '
+        'throughput; same imbalance as total Ly≈q (~1749 cells fail).'
+    ),
+)
+def test_v0_3_total_lci_equals_e() -> None:
```
bl-young: same comment as above I don't think we should name this function after the version number

bedrock/utils/validation/eeio_diagnostics.py
```diff
@@ -465,3 +467,202 @@ def compare_output_from_make_and_use(
...
+    may use χ=1. Failures are loud (``ValueError``), never silent skips.
+
+    Parameters
+    ----------
+    cfg
+        Config to check; defaults to ``get_usa_config()``.
+    """
+    if cfg is None:
+        cfg = get_usa_config()
+    if eeio_year_alignment_precondition_ok(cfg):
```
bl-young: is it necessary to split this across two functions?

bl-young:
> Summary: There are two main failure points.
>
> Ly ≈ q (total) and LCI ≈ E (total) fail because the y value used, y_trade, comes from the Y/trade pipeline, not the IO residual q − A_tot q, so it is not balanced to scaled A_tot/q.

in #458 we adjusted how y_nab is calculated in `derive_cornerstone_y_nab`, are we not using that function here? Is that what you suggest we switch to?

jvendries:
> > Summary: There are two main failure points.
> > Ly ≈ q (total) and LCI ≈ E (total) fail because the y value used, y_trade, comes from the Y/trade pipeline, not the IO residual q − A_tot q, so it is not balanced to scaled A_tot/q.
> 
> in #458 we adjusted how y_nab is calculated in `derive_cornerstone_y_nab`, are we not using that function here? Is that what you suggest we switch to?

Not quite. This is already using `derive_cornerstone_y_nab` from #458 for the domestic checks (`L_dom` / `y_nab`), which is why domestic `Ly≈q` and `LCI≈E` pass.

The total failures come from using `L_tot` with `y_trade` from `derive_cornerstone_Y_and_trade_scaled` rather than using a total (`Adom`+`Aimp`) analogue for `derive_cornerstone_y_nab` (i.e., an IO residual), as the latter function only backcomputes _domestic_ `y_nab`.

What I am suggesting for total is either:

A) validate against `y_tot_bal` = `backcompute_y_from_A_and_q(A=Adom+Aimp, q=scaled_q)` which would be the total check analogous to `derive_cornerstone_y_nab` and be a validation-only construct, or 

B) rebuild/scale `y_trade` so that it balances to `A_tot` and `q`, which would be a broader methodology change.

bedrock/utils/validation/eeio_diagnostics.py
jvendries: Updated precondition flags to match flags after #594 merge

bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
jvendries: Trimmed down tests to: 
A) A loud precondition failure to ensure `model_base_year` is the same as `usa_ghg_data_year`, since we don’t do the chi-matrix year adjustment for this test that useeior does. 
B) One toy LCI pass so wiring/guards fail fast without waiting on @eeio_integration. The live identity check stays in the integration tests.

bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
jvendries: The try block was only there to reset global config after a hard-coded pin. Moved that into a `canonical_usa_config` so the test body is just the assertion path.

bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
jvendries: Changed the tests to use `CANONICAL_USA_CONFIG` (the current snapshot config) rather than a versioned yaml string, and renamed the test to `test_domestic_lci_equals_e`.

bedrock/utils/validation/__tests__/test_eeio_diagnostics.py
jvendries: Same change as in the previous comment, renamed to `test_total_lci_equals_e` and uses the same `canonical_usa_config`.

bedrock/utils/validation/eeio_diagnostics.py
jvendries: Removed the `eeio_year_alignment_precondition_ok` and only kept `assert_eeio_year_alignment_precondition` that raises with the failure reasons.

bl-young (approved):

jvendries:
> **Next steps**
> 
> 1. Either A) validate total `Ly` against `y_tot_bal = q − A_tot q` (parity with `y_nab`), or B) rebuild/scale `ytot/trade` so empirical `y_trade` satisfies `q = A_tot q + y_trade`, then keep the trade-based check. A) is a small validation change; B) is a methodology change to the `Y/trade` pipeline.
> 2. A) Remap BEA to Cornerstone and/or do not drop sectors that influence commodity totals between tables; B) Fix waste disagg bug. A) is a remap of the mapping that may impact other steps in the model build; scope TBD. B) is a limited-scope fix in the waste disaggregation step.

1. Decided to wait for nowcasting project to conclude before updating the Ly and LCI total validations which are currently failing (xfail).
2. Decided to perform additional (limited) scoping to identify specific cause and what it would take to fix.

Issue #576, opened 2026-08-05
Step 1: validation - reconcile each final-use column against published NIPA aggregates


WesIngwersen:
**Nowcast plan Step 1f.** The final-demand counterpart of #537, which does the same for value added.

Plan: [`.claude/plan/nowcast_plan.md`](https://github.com/cornerstone-data/bedrock/blob/nowcast/.claude/plan/nowcast_plan.md) §Step 1f.

## What to build

Validate **each final-demand column** of `derive_initial_Y_pur` against its published NIPA aggregate, per year, as a numeric test rather than an eyeball check.

The **PCE reconciliation to ~1.3%** is the template — that is the standard of proof the other columns should meet.

## Columns and their aggregates

| Code | Column | Aggregate |
|---|---|---|
| F01000 | Personal consumption expenditures | NIPA PCE totals via `BEA_PCEBridge` |
| F02E00 | Nonres. investment in equipment | `BEA_PEQBridge`; reconciles to within 0.22% (#524) |
| F02N00 | Nonres. investment in IP products | NIPA |
| F02R00 / F02S00 | Residential / nonres. structures | NIPA |
| F03000 | Change in private inventories | **NIPA T1.1.5 line 14 = Use `F03000` total exactly** |
| F04000 | Exports | ITA goods+services control (see the trade issues) |
| F06/F07/F10 (12 codes) | Federal/SLG CE, Equip, IP, Structures | NIPA Section 3 |

## Method

Use the three-way comparison from `nowcast_initial_Y_pur_baseline.py` (`4434f7e`), not a two-way one. It separates:

- `ours_minus_bridge` — **our attribution is wrong**
- `bridge_minus_use_sut` — **the source data disagrees**

For 2017 that view showed F01000's gap is entirely ours (the PCE bridge reproduces the Use table to −6 M on 13.3 trillion, 258 of 259 commodities), while F02E00 is mostly ours plus a real +15,031 bridge-vs-Use residual. Conflating the two sends you looking for a bug that isn't there.

## Acceptance

- A per-column, per-year reconciliation table with a stated tolerance for each column.
- Failures are test failures, not notes.
- **Golden file per year**, so later steps cannot silently drift the FD block.

Blocked in practice by the code-space work (#566-#569) — reconciling attribution before the pipeline stops dropping non-NAICS targets means measuring the wrong thing.

WesIngwersen:
[...]

Issue #569, opened 2026-08-05
BEA code space: Phase 1 gate: GHG A/B zero delta with BEA codes present, and remove the NAICS_6 hacks


WesIngwersen:
**P0 — the Phase 1 gate. Nothing in Phase 2 lands before this passes.**

Plan: [`.claude/plan/bea_code_space_cleanup.md`](https://github.com/cornerstone-data/bedrock/blob/plan_bea_code_space/.claude/plan/bea_code_space_cleanup.md)

Depends on #566, #567, #568.

## The gate

1. **GHG A/B returns to zero delta *with the BEA codes present*.** Not "totals match" — see below.
2. The `'531'`, `'23'` and `'92'` entries can be removed from the `NAICS_6` lists in all three target schemas (`BEA_detail_target.yaml`, `Cornerstone_2025_target.yaml`, `CEDA_2025_target.yaml`) **without anything moving**. Those entries exist only to reach BEA codes through a NAICS level; if the Phase 1 fix is real, they are dead weight.

## How to verify — totals are not sufficient

The WIP on `nipa_fd_allocation_fix` (`dde7da4`) preserved the GHG grand total **to the cent** while moving 3920 of 4034 groups and making 15 sectors vanish. A totals check would have passed it.

So the A/B must compare, at minimum:

- **row count** (8695 → 8355 was the regression signal)
- **the set of sectors present** — the 15 that vanished were `230301 230302 233210 233230 233240 233262 2332A0 2332C0 2332D0 233411 233412 2334A0 531HSO 531HST 531ORE`
- **per-sector values** — `531ORE` lost 60.4bn while `GSLGE` gained 18.2bn
- grand total (necessary, nowhere near sufficient)

Use `git stash` A/B as the plan describes.

## Standing metrics to hold or improve

| check | reference |
|---|---|
| `pce_alloc.py` — F01000/F02E00 cells within 1% vs the Use SUT | F02E00 was 14/300 before, **95/306** after the WIP; F01000 237/300 → 238/306; F01000 total −3.20% → −2.98% |
| `nowcast_initial_Y_pur_baseline.py` — per-column totals + the three-way bridge comparison from `4434f7e` | separates "our attribution is wrong" (`ours_minus_bridge`) from "the source data disagrees" (`bridge_minus_use_sut`) |
| `562000` lands exactly | +28,334 |

[...]

Issue #567, opened 2026-08-05
BEA code space: Treat non_naics codes as their own roots in the hierarchy and level machinery


WesIngwersen:
**P0 — the core Phase 1 fix.** Blocked by #566 (the diagnostic answers A-vs-B and therefore what "handle" means here).

Plan: [`.claude/plan/bea_code_space_cleanup.md`](https://github.com/cornerstone-data/bedrock/blob/plan_bea_code_space/.claude/plan/bea_code_space_cleanup.md)

## Goal

Make the hierarchy and level machinery treat a `non_naics` code as **its own root** — no parent, no children, length equal to its own — rather than requiring a row in `NAICS_2017_Crosswalk.csv`.

## Why this shape of fix, and not registration

The alternative — registering BEA codes in the NAICS tables — is what produced the two known failure modes:

- **the dual-target 50/50 split.** While a code remains in `NAICS_2017_Crosswalk`, `non_naics` cannot be its sole route: the melt also produces a target for it (`531HSO` → `531` under the `NAICS_3` default). Housing was observed receiving exactly half its correct value alongside a full-value bogus `531110`.
- **the `'531'`/`'23'`/`'92'` hacks** in the `NAICS_6` lists of all three target schemas, which exist only to reach BEA codes through a NAICS level.

Done properly this should need **no new entries in `NAICS_2017_Crosswalk.csv` at all**.

## Scale of the problem

Of the 406 BEA detail codes in `NAICS_Crosswalk_BEA_2017_Detail.csv`, **210 cannot be reached through any NAICS level**. BEA detail is mostly aggregates (`611A00`, `813100`, `5241XX`, `52A000`, `1111A0`) that are not NAICS codes at all. Every defect chased on 2026-07-29 reduced to the same thing: a BEA detail code that is not a NAICS code cannot be a target, so its value is dropped or renormalised onto whichever sibling happened to be reachable.

| symptom | code(s) | value |
|---|---|---|
| housing collapsed onto a bogus `531110` | `531HSO` `531HST` `531ORE` | 2,026,714 |
| SLG enterprises split across police/prisons/fire | `S00203` | 72,509 |
| waste unattributed | `562000` | 28,334 |
| 27 equipment commodities at zero, 52 overshooting | 29 codes incl. `33329A` `336500` `541300` | 220,644 (18.8% of the PEQ bridge) |
[...]

Issue #566, opened 2026-08-05
BEA code space: Diagnose how the hierarchy and level machinery treat non-NAICS sector codes


WesIngwersen:
**P0 — this is the blocking diagnostic for the whole BEA code-space cleanup.** Nothing else in Phase 1 can be designed until it is answered.

Plan: [`.claude/plan/bea_code_space_cleanup.md`](https://github.com/cornerstone-data/bedrock/blob/plan_bea_code_space/.claude/plan/bea_code_space_cleanup.md)

## Question

How does each of these treat a code that is present in `Sector_Levels` but **absent** from `NAICS_2017_Crosswalk.csv`?

- `bedrock/utils/mapping/naics.py:320`
- `bedrock/utils/mapping/naics.py:370`
- `bedrock/utils/mapping/naics.py:411`
- `bedrock/transform/flowbysector.py:287`

All three `naics.py` functions call `return_naics_crosswalk` and are currently unexamined. They are the **prime suspect** for the GHG breakage described below.

## Why it matters

Two things hang on the answer.

**1. It decides the open A-vs-B design question.**

- **A.** Remove the non-NAICS BEA codes from `Sector_Levels` too, and make the hierarchy functions handle `non_naics` codes as their own roots.
- **B.** Keep both hierarchy tables populated, treat them as the registry of "codes that exist", and let `non_naics` govern only target selection.

A is the coherent end state and matches the intent — *these are not NAICS, stop pretending they are*. B is less invasive. The answer depends entirely on what those three functions do.

**2. It confirms or kills the GHG hypothesis.**

`GHG_national_Cornerstone_2024`, before vs after the WIP on `nipa_fd_allocation_fix` (`dde7da4`): totals identical to the cent, but 8695 → 8355 rows, 3920 of 4034 groups moved, 466 sectors affected, and **15 sectors vanish**:

```
230301 230302 233210 233230 233240 233262 2332A0 2332C0 2332D0 233411 233412 2334A0
531HSO 531HST 531ORE
```

`531ORE` loses 60.4bn, `GSLGE` gains 18.2bn. **The identical grand total is exactly the trap** `About_BEA_IOT_table_valuation_differences.md` warns about — a totals check cannot see a distribution error.

[...]

PR #595, opened 2026-08-05
Salvage the NIPA FD diagnostics and S00300/S00900 fixes from the WIP branch

WesIngwersen:
Salvages the independently-useful work from `nipa_fd_allocation_fix` so it is not stranded behind that branch's `dde7da4` tip, which breaks GHG and must not be merged.

**Targets `nowcast`**, per this project's branching policy. Three commits, cherry-picked with `-x`.

## What's here

| Commit | What |
|---|---|
| `4434f7e` | Compare the final-demand columns against **their bridges**, not just the Use table |
| `ea852b8` | Rename the baseline to `use_sut_detail_PUR` to say what it actually is |
| `e4e24a7` | Land `S00300`/`S00900`, and negate the "Less:" lines that feed `S00900` |

**#576 depends on `4434f7e`.** The three-way comparison is what separates *our attribution is wrong* (`ours_minus_bridge`) from *the source data disagrees* (`bridge_minus_use_sut`). For 2017 it showed F01000's gap is entirely ours — the PCE bridge reproduces the Use table to −6 M on 13.3 trillion, 258 of 259 commodities — while F02E00 is mostly ours plus a real +15,031 bridge-vs-Use residual. Without it you go looking for a bug that isn't there.

`e4e24a7` also adds a general-purpose `negate_flows` clean function to `flowbyclean.py`, for sources that report a quantity as a positive magnitude which the surrounding table subtracts, where the attribution source already carries it as negative. Applied **before** attribution, so ratios distribute a negative total rather than being applied and then flipped — which matters wherever a line splits across sectors.

## Two commits deliberately left behind

**`29b62f0`** — adds `'531'` to the `NAICS_6` lists. That is one of the `'531'`/`'23'`/`'92'` hacks that exist only to reach BEA codes through a NAICS level, and **#569's Phase 1 gate explicitly requires removing them**. Landing it now means landing something Phase 1 undoes. Its durable half (the `9221*`→`GSLGO` retargeting) can be extracted separately if wanted.

**`220682e`** — adds U50505 lines 18 and 41 to a `Line:` list that **no longer exists**. `42f7e59` reverted that design; `FD_IP_equipment` now selects on `Table:` alone and relies on `drop_unassigned`. The underlying gap is real — lines 18 "Special industry machinery, n.e.c." (43,450) and 41 "Electrical equipment, n.e.c." (7,474), together 50,924 or 4.3% of the column — but it belongs to #547, not to a mechanical replay.

## Conflict resolution worth reviewing

`e4e24a7`'s crosswalk hunk conflicted. The incoming side carried four rows — `S00300`, `S00401`, `S00402`, `S00900` — but the commit itself only introduces `S00300` and `S00900`; the other two were already in that branch's tree. **I took only the two.**

`S00401`/`S00402` are currently commented out of `non_naics` in `BEA_detail_target.yaml` (`'23', '92', 'F010' #, 'S00401', 'S00402'`), so on `nowcast` they are reachable by neither route today. That is a live gap, but it is Phase 1's remit (#567) and re-adding crosswalk rows here would prejudge the A-vs-B design question. Left alone deliberately.

## Checks

- `NIPA_FD_2017.yaml` parses
- crosswalk CSV is well-formed — 1292 rows, uniform 5 columns
- `flowbyclean.py` and `nowcast_initial_Y_pur_baseline.py` parse
- `negate_flows` is defined once and referenced by the new activity set

**Not run:** the FBS methods themselves. The standing metric is `pce_alloc.py` — F01000/F02E00 cells within 1% against the Use SUT — and it is worth running before merge, since `e4e24a7` changes the FD_PCE line selection.

## Effect on the WIP branch

After this merges, `nipa_fd_allocation_fix` reduces to `29b62f0`, `220682e` and the experimental `dde7da4`. `dde7da4` was never pushed — the remote tip is `d55e8e0` — so it stays local reference for reproducing the GHG regression (#566), and Phase 1 decides its fate rather than a rebase.


WesIngwersen:
@catherinebirney This is just some work I did that is stranded in a branch that breaks the GHG method so I needed to get this in to the working project master branch, nowcast, to fork other work off of it.

WesIngwersen:
@catherinebirney I will go ahead and merge this one into nowcast since its just some nipa related work so nowcast will be current

PR #597, opened 2026-08-06
Supply/Use match diagnostic: measure Step 1, 2 and 4 progress against the 2017 detail SUT

### LINKED ISSUE #587: Diagnostics: Supply and Use table match visualization, usable as a test in every step
**Cross-cutting.** Not a plan step — a diagnostic that every step is tested with, from Step 1's final-demand columns through Step 6's MUT quartet.

Plan: [`.claude/plan/nowcast_plan.md`](https://github.com/cornerstone-data/bedrock/blob/nowcast/.claude/plan/nowcast_plan.md) §Testing strategy.

## What to build

A visual of the **full Supply table** and the **full Use table** that shows, cell by cell, whether we have input data that matches the reference:

| Colour | Meaning |
|---|---|
| **green** | we have input data and it matches the reference |
| **shade of yellow** | we have input data but the match is imperfect — shade by how far off |
| **white** | no data present to compare |

Applied to the **cells**, and equally to the **row totals and column totals** — the margins are as much the point as the interior.

The result is a single picture of "how much of this table do we actually have, and how good is it", which is the question every step is really asking.

## Build it as two pieces

**1. A comparison engine** — takes a candidate table and a reference table, returns a per-cell status matrix (`match` / `partial` / `absent`) plus the relative error. No plotting. **This is what makes it a test**: it runs headless in CI and returns a pass/fail plus counts, with no image required.

**2. A renderer** over that matrix. This is the diagnostic artifact.

Keeping them separate is what lets the same logic be an assertion in a unit test *and* a picture in a review.

## Design points to settle

**White is currently two different things, and they are not equally bad.**

- the reference has no value there either → genuinely *unknown*, nothing to say
- the reference **has** a value and we produced nothing → a **miss**, which is a failure

Proposal: give the miss its own colour rather than folding it into white, and consider a fourth for the reverse case — we produced a value where the reference has none, i.e. mass in the wrong place. Both are silent under a totals check.

**Tolerance is a parameter, not a constant.** The plan sets different bars per section — PCE reconciles to ~1.3%, PEQ to 0.22%, the Supply identities should be exact. So the green/yellow boundary is per-comparison, and the yellow ramp needs a stated scale rather than an arbitrary one.

**The reference differs by step.** The tool should take a reference selector:

| Step | Reference |
|---|---|
| 1 | 2017 Use SUT final-demand columns |
| 2 | Use SUT value-added rows; NIPA T1.14 / T10305 / T10105 |
| 3 | 2017 Use SUT intermediate block |
| 4 | published 2017 detail Supply table |
| 5 | summary SUT totals (the RAS controls) |
| 6 | the four published 2017 before-redef MUT tables |
| 7 | the 2017 after-redef pair |

## Scale — this needs thought before it is written

BEA 2017 Detail Use is ~402 commodities × ~420 columns, so **~170,000 cells**. That is an image, not an HTML table. Suggestions:

- render via a raster heatmap, not per-cell DOM elements
- offer a **summary-schema view** (71 sectors) as the readable default, with detail available on drill-in
- keep the **row/column total strips** always visible alongside the interior — they are the highest-signal part of the picture and they stay legible at any size

## ⚠️ Why the margins matter as much as the cells

`T014` nets to ~1 economy-wide, and redefinition preserves every total by construction. **A totals check passes on broken data**, which is exactly why this visual is worth building: green interior with yellow column totals, or the reverse, localises the error immediately in a way a scalar cannot.

## Existing precedent to build on

`bedrock/analysis/compare_NIPA_to_IOT/compare.py` already separates **matched-cell disagreement** from **unmatched mass** — the same distinction this needs, and the trade probe showed why it matters (totals within 1% while matched commodities overshot 12% and specials soaked the gap). Start there rather than from scratch.

## Where it lives

Engine in `bedrock/utils/validation/` — alongside the ported `check_balances.py` — so it is importable from tests. Renderer in the analysis/diagnostics path.

## Acceptance

- Callable for any (candidate, reference) pair with a per-comparison tolerance.
- Returns machine-readable counts (green/yellow/white/miss, per cell, per row total, per column total) *and* renders the Supply and Use pictures.
- Used as an actual assertion in at least one step's tests, not only as a report.
- Legible at summary schema; drillable at detail.
- Colour choices readable for colour-vision deficiency — green/yellow/white carries most of the signal in lightness, which helps, but verify rather than assume.

WesIngwersen:
Closes #587.

A diagnostic we run to measure general progress on **Steps 1, 2 and 4**, plus a
report that reads out where the build stands. For any block of the Supply or Use
table it answers the question every step is really asking: *how much of this
table do we actually have, and how good is it* — per cell, and equally per row
total and column total.

## How to run it

```
# render every runnable block and print its report
uv run python -m bedrock.analysis.nowcasting.plots

# regenerate the figures committed alongside the progress report
uv run python -m bedrock.analysis.nowcasting.plots \
    --dpi 110 --out-dir bedrock/analysis/nowcasting/images --no-report
```

`--section <name>` for one block, `--check-palette` for the colour-vision check.

## Where the build stands today

[`bedrock/analysis/nowcasting/progress_report.md`](bedrock/analysis/nowcasting/progress_report.md)
— one document, figures embedded, regenerated per milestone.

| block | step | shape | reference populates | reference total | candidate | coverage | accuracy |
|---|---|---|---:|---:|---|---:|---:|
| `use_fd_detail_sut` | 1 — final demand | 402 × 19 | 1,253 cells | $22.24T | exported | **44.4%** | **32.6%** |
| `use_va_detail_sut` | 2 — value added | 3 × 402 | 1,189 cells | $18.92T | *none yet* | — | — |
| `supply_bridge_detail_sut` | 4 — supply bridge | 402 × 12 | 3,202 cells | $111.28T | *none yet* | — | — |

Those three are the whole of what a published 2017 detail reference supports
outside the two 402 × 402 interiors.

Runnable blocks get their picture and their counts. **Unbuilt blocks get the
reference side**, because the reference is the denominator and it is known
before the step exists — so Step 2's target row totals and Step 4's twelve
bridge codes are stated now, and the bar is set before the build rather than
after. Steps 2 and 4 turn their block on in one line: point
`Section.candidate` at the build output. `Section.runnable` makes the CLI skip
them with a message rather than fail, so the command works today and gets richer
as steps land.

## Layout

Everything lives in one package, [`bedrock/analysis/nowcasting/`](bedrock/analysis/nowcasting/):

```
table_match.py             engine — no plotting
sections.py                the three comparable blocks
plots.py                   renderer + CLI
initial_Y_pur_baseline.py  Step 1 comparison, writes the CSV exports
plan.md                    the build plan
progress_report.md         where the build stands, figures embedded
About_table_match.md       what the first Step 1 run showed
annual_survey_expense_sources.md   Step 3 probe of annual survey data
images/                    the report's figures (tracked)
output/                    working artifacts (untracked)
```

It started split across `utils/validation/` and its `analysis/` subpackage,
which put the engine and its renderer in different directories and sat them
among the EF diagnostics machinery. `utils/validation` is for what
`generate_diagnostics.py` drives; this is nowcast build measurement. The plan
and progress report moved out of `.claude/plan/` for the same reason — that is
session scratch space, and these are project documents that describe the
package they now sit in. `annual_survey_inputs/` folded in too — its Step 3
probe was a package that existed to hold one markdown file. Every file moved as
a git rename, so history follows.

That move also fixed 19 links in the plan that were already dead: they were
written repo-root-relative (`](bedrock/extract/...`), but markdown resolves
relative links against the file, so from `.claude/plan/` they pointed at
`.claude/plan/bedrock/...`.

## Two pieces, deliberately separated

**Engine** — [`table_match.py`](bedrock/analysis/nowcasting/table_match.py).
`compare_tables(candidate, reference, tolerance=...)` → `TableMatch`: status
matrix, relative errors, severity ramp, `Margin` objects for both axes,
`counts()` / `summary()`, and `assert_ok(max_miss=..., min_coverage=...)`. No
plotting, so the same comparison is a picture in review and an assertion in CI.

**Renderer** — [`plots.py`](bedrock/analysis/nowcasting/plots.py).
One `imshow` raster with the margins as strips, which is what lets it draw a
3 × 402 value-added block and a 402 × 402 interior without changing approach.

## Design points the issue asked to settle

**Five statuses, not three.** "No data" is two different things and they are not
equally bad: `MISS` (reference has a value, we produced nothing) and `EXTRA` (we
produced a value the reference has none of) are split out from `ABSENT`. Both
are silent under a totals check and they cancel each other.

**The margins are first-class.** `T014` nets to ~1 economy-wide and redefinition
preserves every total, so a totals check passes on broken data. Row and column
totals go through the *same* `classify()` call as the interior — not a parallel
implementation — so a green interior over a yellow column strip localises an
error no scalar can see.

**One tolerance scale per comparison**, never per row or column: a colour has to
mean the same thing everywhere in a picture. What *is* per-comparison is the bar.
`atol` is half a million dollars throughout, because BEA publishes these tables
rounded to millions.

**Off-frame mass is logged, not dropped.** `TableMatch.residual` totals candidate
mass on labels outside a section's frame and `report()` prints it. The picture
stays on the BEA code space; nothing disappears.

**Colour verified, not assumed.** Every category pair separates by at least
ΔE 27 (CIE76) under simulated protanopia, deuteranopia and tritanopia;
`palette_separation()` re-runs it. The binding pair is not green-vs-yellow —
lightness handles that — but *white against the palest yellow* under tritanopia,
hence a ramp starting at a saturated `#ffd84d`.

## What the Step 1 picture says that the totals do not

```
            absent  match  partial  miss  extra
cells         6385    181      375   697      0
row_totals      22      7      290    83      0
col_totals       0      7        4     8      0

coverage 44.4%   accuracy 32.6%   grand total off by 38.2%
residual outside the frame  6,227,917,314,385  on 14 rows, not drawn
```

Four separable causes, where the grand total alone gives one number:

1. **$6.23T parked on 14 NAICS codes** outside the BEA code space. Two are exact
   identities — `923110` equals the reference's whole `F10C00` column total, and
   `928110` equals `F06C00 + F07C00`. The government columns landed intact, on
   the wrong codes.
2. **Eight whole columns are `miss`**, not scattered partials.
3. **`S00900` right magnitude, wrong sign** (relative error exactly 2.0).
4. **`F01000` amber almost everywhere** — a per-commodity split problem the
   column total understates.

The two axes disagreeing is the point: 373 of 402 row totals outside tolerance
against 12 of 19 column totals. A column total can be right in aggregate while
its commodity split is wrong, and only the row strip shows it.

Longer form: [`About_table_match.md`](bedrock/analysis/nowcasting/About_table_match.md).

## On Git LFS: recommending against it, for now

The figures are committed rather than regenerated on read, so the report renders
on GitHub. **Not** in LFS, and [`.gitattributes`](.gitattributes) records why so
a future reader isn't guessing:

- This is the repo's **first committed image**. `.gitignore` had a blanket
  `*.png` and `git ls-files` finds zero images — so LFS here isn't adopting an
  existing convention, it's introducing one.
- The pack is **~19 MB**. One **124 KB** figure per milestone is a fraction of a
  percent.
- Against that, LFS requires the tool on **every clone and every CI job**, or the
  report renders as pointer text. That cost is paid by everyone, continuously.
- Migrating later **is** easy — only already-committed blobs stay in history, and
  at this size that's nothing. Deferring is cheap and reversible; adopting now is
  neither.

**Revisit if these figures start churning per-commit rather than per-milestone.**
That condition is written into `.gitattributes`.

Two changes keep the committed copies cheap: a `--dpi` option (200 for screen,
110 for the committed copy — 124 KB vs 281 KB, still legible at the 402-row
height), and a `.gitignore` negation scoped to `bedrock/analysis/nowcasting/images/*.png`
**only**, so routine reruns into `*/analysis/output/` cannot accidentally start
committing binaries.

## Known, and out of scope here

Step 1's live candidate does not run: `NIPA_FD_2017.yaml` sets
`attribute_on: ['PrimarySector', 'ActivityProducedBy']`, but the
`retain_activity_columns` plumbing landed in `9668ce7`, was reverted from main in
`42f7e59`, and was **not** part of the `7a04a71` salvage — so the FBS raises
`KeyError: ['ActivityProducedBy'] not in index`. That is Step 1's bug and is left
alone; the section reads the last CSV export meanwhile, and switching back is one
line. Step 1's numbers above are therefore a snapshot of *that run*, not of
`nowcast` HEAD.

No unit tests — this is an analysis script, per review direction. #587's
"used as an assertion in at least one step's tests" therefore stays open;
`assert_ok()` exists and is ready for it.


WesIngwersen:
@bl-young This should generate useable figures and summary stats to check out well we can match the non-core (non industry transactions/make) blocks of the reference tables 

You can see in the report the export col of the Use FD block is totaling missing now.

The Supply table added value block that will house the imports is not yet structured to be produced at all, but thiz viz code will be ready for it. That's where the imports data will go.

bl-young:
Cool this is neat and makes sense. My only comment is on the organization. Would it make sense to place the images and nowcast_progress_report.md inside `bedrock/analysis` instead of inside `.claude`? Similarly `About_nowcast_to_reference_match.md` seems like it would be better off inside bedrock/analysis and not inside `bedrock/.../validation/`

and the `nowcast_*.py` files could maybe go inside a specific nowcast folder as the validation subfolder is fairly cluttered.

WesIngwersen:
> Cool this is neat and makes sense. My only comment is on the organization. Would it make sense to place the images and nowcast_progress_report.md inside `bedrock/analysis` instead of inside `.claude`? Similarly `About_nowcast_to_reference_match.md` seems like it would be better off inside bedrock/analysis and not inside `bedrock/.../validation/`
> 
> and the `nowcast_*.py` files could maybe go inside a specific nowcast folder as the validation subfolder is fairly cluttered.

First regarding the nowcast related modules in `bedrock/util/validation` I was doing that because I thought in essence these are validation checks.  However I see the point that this is a validation to serve this project (or future nowcasting work) and not a validation that needs to be run with every EEIO model, so yes, moving it into `bedrock/analysis ` makes sense as well as moving the scripts and output in `bedrock/util/validation` to the same place. And yes then same can be done for all the progress report and md. summary.  OK will do.

Issue #547, opened 2026-07-28
Step 1A: the PEQ bridge routes household durables to F02E00 that the Use table books under F02R00


WesIngwersen:
**Diagnosed 2026-08-14** with the three-way comparison, now that #638 lets `derive_initial_Y_pur` run end to end. The original defect — line totals split across the wrong commodities — was fixed by #630/#631. What was left is **not** two BEA products disagreeing on a total. It is a scope difference between the PEQ bridge and the Use table, and it is exactly offset by a matching hole in `F02R00`.

## Our attribution is not the problem

`compare_initial_Y_pur_to_bridges_2017` separates the two failure modes:

| column | `ours − bridge` | `bridge − use_sut` |
|---|---:|---:|
| F01000 | +8 $M | −6 $M |
| F02E00 | **−2 $M** | **+15,031 $M** |

We reproduce the PEQ bridge to $2M on $1.17T. The entire gap is bridge-vs-Use.

## The gap is household durables, and F02R00 is short by the same amounts

`bridge − use_sut` for F02E00 is 11 commodities, and each one is missing from our `F02R00` by an **identical** amount:

| commodity | | F02E00 `bridge − Use` | F02R00 `ours − Use` |
|---|---|---:|---:|
| `335220` | household appliances | +11,265 | **−11,265** |
| `314110` | carpets and rugs | +2,235 | **−2,235** |
| `334300` | audio and video equipment | +354 | **−354** |
| `33712N` | institutional furniture | +250 | **−250** |
| `326150` | urethane/foam products | +232 | **−232** |
| `337122`, `337900`, `337121`, `326290`, `335210`, `337215` | furniture, plastics, small appliances | +638 | **−638** |
| | | **+15,031** | **−15,031** |

These are landlord durables — equipment bought for residential rental property. **The PEQ bridge counts them as equipment investment; BEA's Use table books them under residential.** We follow the bridge, so `F02E00` is over-filled and `F02R00` under-filled by the same commodities, cell for cell. The concept is already familiar in this method: PCE has a dedicated `FD_PCE_tenant_landlord_durables` activity set for the same class of goods.

So this is a **routing decision**, not a reconciliation mystery. Nothing is lost — it is in the wrong column.

## What to decide

- [ ] Which target is authoritative for `F02E00` — the PEQ bridge or the Use table?
[...]

PR #601, opened 2026-08-06
Restore retain_activity_columns so the NIPA_FD FBS generates again

WesIngwersen:
Gets the `NIPA_FD_2017` FBS generating again, so Step 1's diagnostic runs live
instead of reading a pinned CSV export. Also refreshes `nowcast` from `main`.

## The bug

`derive_initial_Y_pur(2017)` died with `KeyError: ['ActivityProducedBy'] not in index`.

`NIPA_FD_2017.yaml`'s `FD_PCE_less_nonresident` set asks to attribute on
`attribute_on: ['PrimarySector', 'ActivityProducedBy']`, but `prepare_fbs` drops
the activity columns by default — so the column was gone before the merge that
needed it.

The plumbing that keeps those columns alive landed in `9668ce7`, was reverted
from main in `42f7e59` along with two unrelated commits, and was **not** part of
the `7a04a71` salvage — which re-landed the yaml half that depends on it. The
config has been asking for a column the framework no longer retained.

## The fix

Restores only the substantive part of `9668ce7`: derive `retain_activity_columns`
from `attribute_on` in `load_prepare_attribution_source`, pass it to both
`prepare_fbs` calls. That original commit was mostly quote-style reformatting;
none of that is carried over. `FlowBySector.prepare_fbs` takes `**_kwargs`, which
is why the cached branch needs no signature change.

**Opt-in by construction.** It only fires when `attribute_on` names an activity
column, and `NIPA_FD_2017.yaml` is the only config in the repo that does — the
other 48 `attribute_on` declarations name sector or flow columns and are
untouched.

## Step 1 now runs live

`Section.candidate` moves from the pinned export to `derive_initial_Y_pur`:

| | export | live |
|---|---:|---:|
| match | 181 | 182 |
| partial | 375 | 374 |
| grand total error | −38.235% | −40.042% |

`initial_Y_pur_exported_candidate` stays. Reading a pinned export is the only way
to put a past run beside a current one, and it is the fallback if the FBS breaks
again.

The report and its figure are regenerated from the live run.

## `nowcast` refreshed from `main`

`nowcast` was 8 behind; it is now level, and this branch carries that merge.
Neither merge had a textual conflict, and the merged state produces **identical**
Step 1 numbers to the pre-merge live run.

Worth recording because it was the reason for checking: **#561 deletes
`use_cornerstone_2026_model_schema`**, which `nowcast.py`'s
`_force_cornerstone_model_schema` read and wrote — that would have broken
`derive_initial_Y_pur` on merge. It doesn't, because #561 deleted the context
manager in the same commit. The only surviving references are stale recorded
config dumps under `analysis/reconciling_data_years/output/`, which are outputs
rather than live code.

**On the mypy question:** `flowby.py` is byte-identical on `main` and `nowcast`,
so there was no fix in main to pick up. Repo-wide mypy is unchanged before and
after the merge — 4 errors, all in `utils/config/settings.py`, all
`resource.RLIMIT_AS` / `setrlimit`, which don't exist on Windows. They're a
platform artifact of running mypy locally, not a real failure. (The 8
"unused type: ignore" errors I mentioned earlier only appear when checking
`flowby.py` in isolation; they vanish in the full-package run.)

## What now actually limits Step 1

The caveats section is rewritten. The FBS-does-not-run entry is replaced by the
two real constraints, both **#547 work item 2**, neither addressed here:

- **Only one of seven PCE activity sets attributes on its own bridge rows.**
  `9668ce7` set `attribute_on` on six PCE sets and the equipment set; the
  `7a04a71` salvage re-landed it on `FD_PCE_less_nonresident` alone. So the
  per-commodity split fix that commit reported (F01000 18/297 → 189/293 cells
  within 1%) is only partly in place — consistent with `F01000` still being
  amber almost everywhere in the picture.
- **F02E00's equipment names are unresolved.** `U50505` carries 45 activity
  names against `BEA_PEQBridge`'s 27, only 26 matching exactly, so NIPA rows
  drop rather than attribute.

`S00900`'s sign error is now known to be **current** rather than an artefact of a
stale export, since this run postdates `7a04a71`'s `negate_flows`.

## Verification

- 105 transform + extract tests pass. One further error is an expired GCS
  credential, not this change.
- `ruff`, `black` clean; repo-wide mypy unchanged.
- Step 1 renders end to end; the committed figure reproduces byte-for-byte from
  the command the report documents.
- No broken relative links in the package.


bedrock/transform/flowby.py
```diff
@@ -1080,6 +1080,16 @@ def load_prepare_attribution_source(
         else:
             ((name, config),) = attribution_source.items()
 
+        # `attribute_on` may name activity columns, which prepare_fbs drops by
+        # default. Keep them when they are about to be merged on, so an
+        # attribution source can be matched per activity rather than only per
+        # sector -- e.g. splitting a NIPA PCE line by that line's own rows of
+        # the PCE bridge, instead of by the bridge summed over every line.
+        attribute_on = (attribution_config or self.config).get('attribute_on') or []
+        retain_activity_columns = any(
```
WesIngwersen, line 1089: @catherinebirney Just an FYI. This had been lost in the nowcast branch in cherrypicking earlier work and the FBS was breaking. This is to retain the data in these columns for later use which is done only in this method. A quick blast radius check showed no impact on other methods. Now at least the 2017 FBS will run in nowcast.

PR #608, opened 2026-08-07
chore(config): pair waste-disagg test configs, load Cornerstone GHG

### LINKED ISSUE #604: Stop standard CI regenerating CEDA FBS for waste-disagg B tests
## Summary

Standard CI (`pytest` with `not eeio_integration`) spends ~4–5 minutes regenerating `GHG_national_CEDA_2023` while building B for unmarked waste-disagg pipeline tests. Those tests only need sector-column shape / waste vs non-waste behavior; which GHG FBS fills E does not matter. This is a relic from #594 and #553 

## Goal

Set `use_cornerstone_ghg_model: true` on both configs used by the B fixtures so E loads the pre-built Cornerstone FBS parquet from GCS:

- `test_usa_config_waste_disagg` (`disagg_B`)
- `2025_usa_cornerstone_taxonomy_and_B_transformation` (`baseline_B`)

Both must flip; updating only the waste-disagg config leaves the baseline fixture on the CEDA regen path.

This will save 4-5 minutes on each CI

bl-young:
cc:
Closes: #604

## What changed? Why?

Waste-disagg pipeline integration tests use a paired on/off config pair (`test_usa_config_waste_disagg` / `test_usa_config_waste_disagg_off`) with the same years and `use_cornerstone_ghg_model: True`. The off config replaces `2025_usa_cornerstone_taxonomy_and_B_transformation`, which only served as that baseline.

Both fixtures load the pre-built Cornerstone GHG FBS parquet from GCS instead of regenerating `GHG_national_CEDA_2023` via `getFlowBySector`, which was costing ~4–5 minutes in the standard CI `test` job. 

## Testing

```powershell
uv run pytest bedrock/transform/eeio/__tests__/test_waste_disagg_pipeline_integration.py -m "not eeio_integration" -v
```


jvendries:
Tested locally with 21 tests passing, 2 deselected, 1 xfail

jvendries (approved):

Issue #568, opened 2026-08-05
BEA code space: Allow SectorSourceName to express a non-NAICS schema


WesIngwersen:
**P0 — Phase 1, third piece.** Can be decided in parallel with #567; both are gated by the same Phase 1 verification.

Plan: [`.claude/plan/bea_code_space_cleanup.md`](https://github.com/cornerstone-data/bedrock/blob/plan_bea_code_space/.claude/plan/bea_code_space_cleanup.md)

## The problem, stated plainly

The pipeline has **no way to express "these sectors are BEA detail codes"**, so a correct crosswalk is forced to misdeclare itself.

`NAICS_Crosswalk_BEA_2017_Detail_identity.csv` maps 406 BEA codes to themselves. Every row must still say `SectorSourceName = NAICS_2017_Code` or the loader drops it — because `flowbyactivity.py` (~line 520) parses `SectorSourceName` as `NAICS_<year>_Code` and then filters rows to that literal.

A sibling guard at `flowbyactivity.py` (~line 486) selects the whole mapping path on `"NAICS" in activity_schema`, so non-NAICS sources take the crosswalk path **by accident, not by design**.

## Decision required

Can `SectorSourceName` express a non-NAICS schema, with the schema carried through the loader?

The plan's own view: *"A real fix probably means allowing a non-NAICS `SectorSourceName` and carrying the schema through, rather than adding more codes to the NAICS tables."*

Scope the change to the two guards above plus anything downstream that branches on the parsed value. This is the piece that makes the identity crosswalk honest rather than merely working.

## Acceptance

- An activity-to-sector crosswalk whose sectors are BEA 2017 detail codes can declare that fact, and load.
- The `"NAICS" in activity_schema` branch selects the mapping path deliberately, with non-NAICS sources taking a path chosen for them.
- No behaviour change for existing NAICS-declaring crosswalks — verified per column and per sector, not on totals.

## Related

- #566 (diagnostic), #567 (hierarchy fix) — same Phase 1.
- Once this lands, the identity crosswalk in `dde7da4` can stop declaring `NAICS_2017_Code`.

Issue #528, opened 2026-07-23
Step 1B: Trade FBS - implement the trade FBAs and FBS methods


WesIngwersen:
**Rescoped 2026-08-14 (#523).** Trade is its own FBS. Implementation for #526.

## Built

- `Census_USATrade` and `BEA_ITA`/`BEA_IEA` FBAs (#617)
- `Trade_Exports_2017`, `Trade_Imports_2017` FBS methods (#618)
- wired into the nowcast `Y` `F04000` and Supply `MCIF` (#622)
- 2017 F040/MCIF scorecard and IEA leaf crosswalk (#623)
- Trade FBS on shared `BEA_detail_commodity_target.yaml` (#638)
- nowcast ITA G+S scale on F040/MCIF overlay (#642); FBS parquet stays unscaled
- #618 README/YAML follow-ups: drop `primary_action_type`; document zero-weight concentration (#642)

## Remaining

- [ ] FAS→PUR commodity conversion
- [x] `MDTY` / `MADJ` (#579)
- [ ] `S00300` hold-from-Supply (#606)
- [x] 2018–2024 Trade methods
- [ ] 2017 scorecard is **FAIL + inventory** against the #557 bars — close the gap (ITA control matches ITA; SUT national % is a separate benchmark)

bl-young:
Flagging for this issue that the USEEIO-based approach was integrated into flowsa here: https://github.com/cornerstone-data/flowsa/compare/master...cornerstone-data:flowsa:imports

These loaders are not in bedrock from our original port.

WesIngwersen:
**`S00900` is not a trade-data problem. It needs no extract at all — it is determined by the supply identity, and its other half is already built.**

The plan lists `S00900` (~204B) as the single biggest hole in `F04000`, inside the "27.5% of Use `F04000` value has zero extract" figure, and asks for an explicit specials rule. This answers it for `S00900`: **derive it, don't source it.**

Found while checking whether the Section 3 Rest-of-the-World lines reconcile against it. They don't, and they shouldn't — see the end.

## The whole commodity, 2017 detail SUT

`S00900` has **zero intermediate use** — nothing across any of the 402 industries. Two final-demand entries and nothing else. It *is* supplied, though, which is easy to miss:

```
SUPPLY  produced by S00600 (federal nondefense)     3,468
        imports MCIF                                  -26
        total supply T013 / T016                    3,441

USE     intermediate, all 402 industries                0
        PCE      F01000                           -200,997
[...]

PR #618, opened 2026-08-09
feat(transform): add 2017 Trade FBS methods and Detail Crosswalks

bl-young:
cc:
Related: #528

## What changed? Why?

Maps national Census goods and BEA IntlServTrade services to BEA 2017 Detail commodities.

Out of scope: nowcast Y `F04000` / Supply `MCIF` wiring, and ITA G+S scale.

**Crosswalks** under `activitytosectormapping/`:

- `NAICS_Crosswalk_Census_USATrade.csv` — Census NAICS-6 → BEA 2017 Detail (`SectorType` C). Concordance content from USEEIO [`Census_to_useeio2_sector_concordance.csv`](https://github.com/USEPA/USEEIO/blob/master/import_emission_factors/concordances/Census_to_useeio2_sector_concordance.csv).

- Empty-sector NAICS `980000` is omitted. Census NAICS `331314` maps to industry code `331314`, which is not a 2017 SUT commodity (commodities are `331313` / `33131B`).

- `NAICS_Crosswalk_BEA_IEA.csv` — IntlServTrade `TypeOfService` → BEA 2017 Detail. Concordance content from USEEIO [`BEA_service_to_useeio2_sector_concordance.csv`](https://github.com/USEPA/USEEIO/blob/master/import_emission_factors/concordances/BEA_service_to_useeio2_sector_concordance.csv) (~10 API types).

- Hierarchy totals such as `AllTypesOfService` are not mapped. `GovtGoodsAndServicesNie` includes industry code `S00101` (federal electric utilities), which is not a 2017 SUT commodity.

- `SectorSourceName` is `NAICS_2017_Code` because the loader filters on that literal (#568).

**FBS methods** `transform/trade/Trade_Exports_2017.yaml` and `Trade_Imports_2017.yaml`:

- Exports: Census `ALL_VAL_YR` + BEA Exports → `F04000` on `SectorConsumedBy` after aggregation.

- Imports: Census `GEN_CIF_YR` + BEA Imports for the MCIF-family import vector.

- 1:m Crosswalk rows use `attribution_method: proportional` against 2017 Use SUT `F04000` (exports) and Supply `MCIF` (imports).

- SUT weight sources set `activity_schema: NAICS_2017_Code` so Detail commodity codes are not run through `BEA_2017_Detail` (Detail→NAICS). `Trade_detail_passthrough.yaml` keeps them from collapsing.

- Calling sets use `primary_action_type: Produced` because the trade FBAs are `TECHNOSPHERE_FLOW` with empty `ActivityConsumedBy`.

- Import weights use `primary_action_type: Consumed` (`MCIF` is `ActivityProducedBy`, commodity is `ActivityConsumedBy`).

- Duties and charges stay on the Census FBA `FlowName`.

**Target include** is Trade-local `Trade_detail_passthrough.yaml` (`non_naics` identity so Detail `Sector` strings are not NAICS-collapsed).

- Not a shared BEA schema, and not `Cornerstone_2025_target.yaml` / `BEA_detail_target.yaml` (those select NAICS leaves; NIPA FD Crosswalks emit NAICS).

- Swap when #567 / #568 land.

**Probes / specials**

- Specials policy is in `transform/trade/README.md`.

- Removes `probe_2017_trade_detail` and `probe_2017_trade_compare`. Method check is `generateFlowBySector`.

- `probe_2017_trade_totals` stays as the FBA national-totals check.

## Testing

Requires local 2017 `Census_USATrade` and `BEA_IEA` FBAs.

```powershell
uv run python -c "from bedrock.transform.flowbysector import FlowBySector; e=FlowBySector.generateFlowBySector('Trade_Exports_2017', download_sources_ok=False); i=FlowBySector.generateFlowBySector('Trade_Imports_2017', download_sources_ok=False); print(e['FlowAmount'].sum()/1e6, i['FlowAmount'].sum()/1e6, e['SectorConsumedBy'].unique().tolist(), len(e), len(i))"
```

2017 smoke:

- Exports: 1,963,339 million USD, 294 rows, `SectorConsumedBy` is `F04000`.

- Imports: 2,696,418 million USD, 283 rows.

- Generate warns unmapped Census `980000` and IEA `TypeOfService` codes outside the thin Crosswalk.

- SUT cells with no FBS mass include `S00300` (imports) and `S00900` / `533000` (exports).

- Extract-only code `331314` is not a 2017 SUT commodity.


bl-young:
@WesIngwersen - note that this is really just the initial setup - the mappings still need further revision and I assume will be impacted by the change to the schema.

WesIngwersen:
>(MCIF is ActivityProducedBy, commodity is ActivityConsumedBy).

I understand neither are intuittive but I think the reverse is a cleaner setup .. when all the FD columns are ConsumedBy for final demand so commodites are ProducedBy

WesIngwersen:
>Duties and charges stay on the Census FBA FlowName

The duties will need to be with a separate col of FD, i think also with the commodity as  ProducedBy and `MDTY` as the ConsumedBy. Whether you do this in the same or a separate method is up to you.

The charges have to be explored more against the BEA data. I think they can stay associated with the commodity and then we can reallocate them to the trade sectors in nowcast.py

WesIngwersen (commented):
See my comments and decide if you want to modify for some or all of those now or later - up to you.

WesIngwersen (commented):
Reviewed by pulling the branch into a scratch worktree, loading both method configs through the real YAML loader, and running the actual 2017 Use/Supply weight vectors against the new crosswalks. **No blocking bugs** — the mechanics hold up better than the description claims. Details below.

## What I verified

**Crosswalks are mechanically clean.** 415 Census rows / 62 IEA rows; no duplicate `(Activity, Sector)` pairs, single `ActivitySourceName` each, all `SectorSourceName = NAICS_2017_Code`. Every one of the 246 Census and 62 IEA target sectors appears in the 404-entry `non_naics` list — so nothing gets silently NaN'd out of the target spec. The `non_naics` list itself has no duplicates.

**No mass is lost to attribution.** Only 1 of 414 Census activities is 1:m (`11211X`); all 10 IEA activities are. I checked every 1:m group's denominator against the real F04000 and MCIF vectors — all non-zero, so nothing falls into the silent `denominator == 0` drop path at `bedrock/transform/flowby.py:1256-1259`. Weights are all non-negative except MCIF `S00900` (−26M), which isn't in any 1:m group. The smoke totals reconcile: exports 1.963T ≈ 1.546T goods FAS + ~0.5T mapped services; imports 2.696T ≈ 2.407T goods CIF + ~0.264T mapped services.

## Worth changing

**1. The `primary_action_type` rationale in the README is wrong, and the setting is a no-op.** `selection_fields: {ActivityConsumedBy: {'F04000': ''}}` is the *mapping* form — it selects and renames to `''`, so the FD/MCIF label is blanked before sector mapping. Measured: the exports weight comes out 341 rows with `SectorConsumedBy` NaN throughout and `PrimarySector` = commodity; the imports weight comes out 298 rows with `SectorProducedBy` NaN throughout. So "Import weights set `primary_action_type: Consumed` because Supply stores `MCIF` on `ActivityProducedBy`" describes state that no longer exists by the time `add_primary_secondary_columns` runs. Either drop the setting, or keep it as defensive and fix the comment — and if you keep it, add `primary_action_type: Produced` to `use_f040` for symmetry so the two anchors read the same way.

**2. Losing the commodity-level check.** The analysis README now says "Commodity-level scoring is `Trade_Exports_2017` / `Trade_Imports_2017` FBS vs Use `F04000` / Supply `MCIF`" in the same commit that deletes the only code doing that scoring (`probe_2017_trade_detail.py`'s Pearson/coverage, plus the `compare()` probe). `generateFlowBySector` proves the method runs, not that the vectors agree. I'd repoint the detail probe at the FBS output rather than delete it — otherwise that sentence in the README has nothing behind it. (Nothing dangles, though: no stale references to either probe survive on the branch.)

**3. Several crosswalk targets get zero weight, so mass concentrates harder than the crosswalk implies.** Worth a README line, since reading the CSV alone is misleading:

- `MaintenanceAndRepairNie` imports — 811100/811200/811300 have no MCIF cell, so 100% lands on 811400
- `Insurance` — 524113 and 524200 are absent from both, so 100% lands on 5241XX
- `GovtGoodsAndServicesNie` — S00101 and S00102 absent from both, so all $710M exports / $289M imports land entirely on 491000 (postal)

Also partial for `ProfMgmtConsult`, `TechTradeRelatedOth`, `PersonalCulturalAndRecreational`. The README notes S00101 is industry-only but not that the consequence is total concentration onto postal.

## Nits

- `SectorType: C` is a value no other crosswalk in the repo uses (the existing convention is `I` or blank) and nothing branches on it — it's carried and dropped. Fine, but state "C = commodity" in the README so nobody assumes it's enforced.
- The attribution anchors hardcode `year: 2017` rather than `*trade_year`. When these get templated per year like `NIPA_FD_<year>`, that's a third spot to remember.
- `target_naics_year: 2017` is set in both the passthrough include and each method. Redundant, harmless.
- The 404-code passthrough list is hand-maintained and duplicates content derivable from `NAICS_Crosswalk_BEA_2017_Detail.csv`. Acceptable for something explicitly temporary (#567 / #568), but a comment on how it was generated would help whoever refreshes it.
- Include resolution works (`Trade_detail_passthrough` → `trade_detail` → `trade`), but the search order in `bedrock/utils/config/flowsa_yaml.py:38-48` puts `transform/common` *before* the method's own folder — so never add a same-named file there.
- No automated coverage; `bedrock/transform/__tests__/test_fbs.py` only pins the Cornerstone GHG method. Not a blocker given repo convention, just noting these methods are validated by hand only.

bl-young:
ok thanks I will address these in a follow up PR

PR #625, opened 2026-08-10
Implement config updates for electricity disagg

jvendries:
cc:
Closes:

## What changed? Why?

- **Add v0.3.1 electricity configs** (`footing` → `reallocation` → `disaggregation` → `mixed_units`, margins off, `apply_io_year_adjustments`, pin `00524c3c…`) so the electricity disagg chain can run on the current Cornerstone methodology instead of the original v0.2 configs.
- **Retarget electricity diagnostics** to those configs and `vs_v0_3` outputs so analyses compare against the v0.3.1 electricity footing rather than stale v0.2 sheets and labels.
- **Port the EIA-anchored T&D markup counterfactual** onto the v0.3 configs (plus `cache_reset` clears for realloc/disagg flags) so that diagnostics-only alternative split stays available after the footing move.
- **Fix cornerstone inflation helpers for elec × `apply_io`** (industry/commodity PI axes at 407, fail-loud A inflate, diagnostics cache clears) so inflated A/q no longer crashes on the 405↔407 mismatch.
- **Regenerate diagnostics artifacts** (new Google Sheet IDs in `manifest.yaml`, BLy/EF PNGs) and stop tracking `local_data/*.xlsx` so sheet exports stay local while plots and sheet-based analyses use the new v0.3.1 workbooks.

## Testing

- `test_local_data.py` updated to use v0.3 config names and the new `local_workbooks` import; a new `test_v03_electricity_configs_load` test verifies that the four new YAML configs load with the expected field values including the v0.3.1 snapshot SHA.
- `test_manifest.py` updated with a new `test_load_manifest_can_skip_sheet_id_check` test confirming the footing config resolves correctly without sheet IDs, and `test_expectations_include_methodology_and_sha` verifying that `RunExpectation` entries carry the v0.3.1 SHA and correct methodology flags including `cornerstone_industry_avg_margins`, `apply_io_year_adjustments`, `model_base_year`, and `usa_ghg_data_year`.
- `test_net_change.py` label assertions updated to `'Cornerstone v0.3.1 electricity footing'`.
- `test_inflation_helpers_cornerstone.py` expanded with tests covering industry and commodity price ratio axis correctness under `apply_io_year_adjustments` with and without electricity disaggregation, a smoke test for `derive_cornerstone_Aq_scaled` on the v0.3 elec config, and a fail-loud test for `inflate_cornerstone_A_matrix_with_commodity_pi` on mismatched labels.


jvendries:
* **#632** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/632?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#625** <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/625?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cornerstone-data/bedrock/625?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bedrock/analysis/electricity_disagg_diagnostics/README.md
```diff
@@ -179,6 +190,10 @@ python -m bedrock.analysis.electricity_disagg_diagnostics.full_trace.decompose_d
 Documents year handling for E / B / A / q / L / D / N under mixed units, and
 probes a single-year 2017 attempt (blockers + proxies).
 
+**Carve-out:** `year_alignment_bly_e.py` still uses the v0.2 mixed-units config
+(`MIXED_CONFIG = 2025_usa_cornerstone_v0_2_electricity_mixed_units`), not the
+v0.3.1 suite chain.
```
bl-young: why?

bl-young (commented):
- With #548 we don't need to specify the snapshot for diagnostic comparison within the config itself. Given that, is the `2025_usa_cornerstone_v0_3_electricity_footing` needed? I'm not sure the margins flag should really matter in most cases.
- Would this enable us to drop the v0_2 electricity configs? I'm not sure we should need those for anything.
- There was a lot of change in `eia_anchored_td_markup_counterfactual.py` that I'm not seeing in the PR description - can you describe what that's for?

>Fix cornerstone inflation helpers for elec × apply_io (industry/commodity PI axes at 407, fail-loud A inflate, diagnostics cache clears) so inflated A/q no longer crashes on the 405↔407 mismatch. 

- Was this ^^ a bug that we had before? I'm not sure I understand what this means.

bl-young:
**Regarding the test suite:**

Most of the new tests don’t clear the repo bar (real failure mode a careful read of the diff would miss). Suggested cut:

**Keep**
- `test_industry_price_ratio_apply_io_plus_elec_is_industry_elec_indexed` — this is the actual axis/405↔407 regression. Fine to also assert the commodity ratio is `CORNERSTONE_COMMODITIES_ELEC` in the same test. Drop redundant `len == 407` / per-sector `in index` checks implied by list equality.
- `test_summary_commodity_price_ratio_cache_ignores_call_style` — named historical-bug pin for the positional-vs-kwargs `@cache` split.

**Drop**
- `test_vnorm_commodity_price_ratio_apply_io_plus_elec_is_commodity_elec` — merge into the industry-axis test if you want both axes; identity@year is already covered
- `test_aq_scaled_smoke_apply_io_plus_elec` — smoke; axis pin already covers the crash mode
- `test_industry_price_ratio_apply_io_no_elec_stays_405` — unchanged non-elec path
- `test_industry_price_ratio_v0_2_elec_is_commodity_elec_indexed` — untouched `apply_io=False` leg
- `test_inflate_a_with_commodity_pi_fails_loud_on_missing_labels` — defensive `KeyError` on a synthetic label the pipeline never produces
- Extra reverse roundtrip in `test_adjust_summary_A_dollar_year_roundtrip` — redundant with the cache call-style test
- `test_v03_electricity_configs_load` — asserts YAML fields the YAML declares
- `test_load_manifest_can_skip_sheet_id_check` / `test_expectations_include_methodology_and_sha` — “did we add the kwarg / hardcode the defaults?”; trust code review of `_methodology_expectation` and the YAMLs

Net: ~8 new tests → ~2.

bl-young:
Have you done an assessment of the diagnostics here when stacked on v03 compared to when it was stacked on v02? Do we expect much real change in the story (my recollection is no). (or are you holding off for this kind of review until you get to the other updates?)

jvendries:
> * With #548 we don't need to specify the snapshot for diagnostic comparison within the config itself. Given that, is the `2025_usa_cornerstone_v0_3_electricity_footing` needed? I'm not sure the margins flag should really matter in most cases.

We haven't done a specific electricity analysis for margins on/off yet, so I don't know how much it would change results. This is the main reason why the v0.3 electricity footing is there, as production v0.3 config is margins on. Determining the effect of margins on electricity disagg will be a follow up step after which we can drop the electricity footing config if it is not needed.

> * Would this enable us to drop the v0_2 electricity configs? I'm not sure we should need those for anything.

Made the edits required to drop the v0.2 configs (updated `year_alignment_bly_e.py`, removed tests that referenced those configs) and then dropped them.


> * There was a lot of change in eia_anchored_td_markup_counterfactual.py that I'm not seeing in the PR description - can you describe what that's for?

I added text to the PR description to describe the changes in the `eia_anchored_td_markup_counterfactual.py`. Note that it’s still diagnostics-only (no production EEIO change).

>>Fix cornerstone inflation helpers for elec × apply_io (industry/commodity PI axes at 407, fail-loud A inflate, diagnostics cache clears) so inflated A/q no longer crashes on the 405↔407 mismatch.

> * Was this ^^ a bug that we had before? I'm not sure I understand what this means.

This was not a bug under v0.2 electricity configs, but the code had to be adjusted for v0.3 electricity configs. To elaborate: v0.2 electricity never used `apply_io_year_adjustments`, so those configs never hit this failure; instead they used the legacy (`apply_io=False`) branch.

`apply_io` alone on production v0.3 also worked, as the dimensions of the matrices never changed (stayed at 405). The bug is specifically `apply_io` + electricity disagg (407 sector matrices). The `apply_io` inflate helpers weren’t correctly wired for the expanded electricity sectors (and silent fill_value=1.0 / cached 405 PI made it worse). That combo didn’t exist until this PR where the v0.3 electricity configs suite put both flags on.

bedrock/analysis/electricity_disagg_diagnostics/README.md
jvendries: Updated the `year_alignment_bly_e.py` analysis to use v0.3. Quick summary of the v0.2 vs v0.3 output:

### v0.2 → v0.3 comparison

| | v0.2 baseline | v0.3 baseline | Δ |
|---|---:|---:|---:|
| Years | model/ghg **2023**, io 2022 | model/ghg **2024**, io 2022, `apply_io` | — |
| National BLy/E | 1.014 | **1.071** | +5.6% |
| Elec BLy/E | 1.121 | **1.372** | +22.3% |
| Elec x/q median | 1.119 | **0.915** | −18.2% |
| Elec E (Mt) | 1471 | 1475 | +0.3% |

| | v0.2 2017 attempt | v0.3 2017 attempt | Δ |
|---|---:|---:|---:|
| National BLy/E | 0.9965 | 1.001 | +0.5% |
| Elec BLy/E | 0.9867 | **1.000** | +1.3% |
| Elec E / c_col | same | same | 0% |

**Takeaway:** Story is the same (identities hold; `BLy ≠ E` under production years because of the `q`–`x` wedge). Magnitudes move: v0.3 baseline opens the elec BLy/E gap (~1.12 → ~1.37), consistent with the larger `q`–`x` mismatch under `apply_io` / 2024. The 2017 freeze probe is nearly unchanged and, under the v0.3 YAML base, elec BLy/E lands at 1.0.

jvendries:
> **Regarding the test suite:**
> 
> Most of the new tests don’t clear the repo bar (real failure mode a careful read of the diff would miss). Suggested cut:
> ...
> Net: ~8 new tests → ~2.

Agreed. Removed the suggested unneeded tests.

jvendries:
> Have you done an assessment of the diagnostics here when stacked on v03 compared to when it was stacked on v02? Do we expect much real change in the story (my recollection is no). (or are you holding off for this kind of review until you get to the other updates?)

I did a quick side-by-side of the old v0.2 diagnostics sheets vs the new v0.3.1 suite. The story doesn't change: reallocation is small, the 3-way split drives almost all BLy change, and mixed units leaves totals flat. Magnitudes do change: electricity-block FINAL BLy is ~23% higher and the realloc→3-way jump is ~2.8× larger under v0.3.1 (apply_io + 2024 years), consistent with a larger q–x wedge in the year-alignment probe.

Note: a pending scaling fix under `apply_io` (upcoming PR) would shift those post–3-way magnitudes again (~10% order on the electricity scale path) without changing which cascade step dominates. Treat current levels as temporary until that fix is implemented.

### v0.2 vs v0.3.1 electricity diagnostics (BLy sheets)

**Story:** unchanged — realloc small → 3-way drives almost all BLy change → mixed ≈ 0.

| Metric | v0.2 | v0.3.1 | Δ |
|---|---:|---:|---|
| National footing BLy (Mt) | 4732 | 4655 | −1.6% |
| National mixed BLy (Mt) | 4940 | 5273 | +6.7% |
| Elec footing BLy (Mt) | 1420 | 1399 | −1.4% |
| Elec mixed BLy (Mt) | 1650 | 2024 | +22.7% |
| realloc→3-way elec (Mt) | +225 | +621 | 2.8× |
| 3-way→mixed elec (Mt) | ~0 | ~0 | same |
| Year-align elec BLy/E | 1.121 | 1.372 | +22% |


**Legend**

| Row | Meaning |
|---|---|
| **National footing BLy** | Total attributed production `BLy` (all sectors) on the pre-disagg footing sheet. |
| **National mixed BLy** | Same total at the FINAL mixed-units step (end of the cascade). |
| **Elec footing BLy** | Electricity-block `BLy` only (`221100` on footing; G/T/D children after split). |
| **Elec mixed BLy** | Electricity-block `BLy` at FINAL (sum of `221110`/`221121`/`221122`). |
| **realloc→3-way elec** | Change in electricity-block `BLy` when going from reallocation → 3-way monetary split (the big step). |
| **3-way→mixed elec** | Change in electricity-block `BLy` when converting gen to physical units (should be ~0 on totals). |
| **Year-align elec BLy/E** | From the mixed-units year-alignment probe: electricity-block `BLy / E` (1 = attributed matches inventory). |

bl-young:
> Determining the effect of margins on electricity disagg will be a follow up step after which we can drop the electricity footing config if it is not needed.

I think the margins should only be relevant as a post-pipeline additional step. So I don't believe it matters. But its not a big deal.

>  the realloc→3-way jump is ~2.8× larger under v0.3.1 (apply_io + 2024 years), consistent with a larger q–x wedge in the year-alignment probe.

This feels quite large and hard to explain, maybe i'm not tracking. I guess I can look at the next PR too

jvendries:
### Merge activity

* **Aug 14, 2:52 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/625).
* **Aug 14, 2:53 PM UTC**: @jvendries merged this pull request with [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/625).
* **Aug 14, 2:53 PM UTC**: [Graphite](https://app.graphite.com/github/pr/cornerstone-data/bedrock/625) couldn't merge this PR because it failed for an unknown reason.

Issue #620, opened 2026-08-09
Step 4c: validate the annual transport margin trend against BTS_TSA and Freight Facts and Figures


WesIngwersen:
**Priority: medium.** Parent: #571. Follows #611 (Step 4c-2), which built the annual transport margin.

#611 derives an annual `TRANS` column as **shape from FAF ton-miles, level from a control total**, and the control total turned out to carry most of the movement — the volume index reaches 1.036 by 2022 while the nominal level reaches 1.483. That 1.43x repricing is the single largest thing the construction does, and **nothing external has been used to check it.**

Two BTS sources can, and one is already half-built in bedrock.

## Source 1 — `BTS_TSA`, Transportation Satellite Accounts

Already an FBA: [`bedrock/extract/bts/BTS_TSA.py`](https://github.com/cornerstone-data/bedrock/blob/main/bedrock/extract/bts/BTS_TSA.py) + `.yaml`, with `NAICS_Crosswalk_BTS_TSA.csv`. Pulls the TSA Make and Use tables from `data.bts.gov`, and `tsa_parse` already separates **in-house** production of transportation services from **for-hire**.

Why it is the right check: the #611 control total scales each mode's *whole* gross output by its 2017 ratio of margin-given-up to output. TSA gives for-hire output **by mode** directly, which is the universe the transport margin actually draws on, so it tests the assumption without inheriting it.

Two things to sort out:
- The FBA yaml lists **2012-2019 only**, and the nowcast window runs to 2024. Check whether the `5yqg-88j3` dataset now carries later years and extend `years:`.
- The crosswalk is `NAICS_2012_Code`. The margin chain is on 2017 detail.

The in-house series is worth a look in its own right: own-account transportation is *not* margin (it sits in the producing industry's own costs), so a shift between in-house and for-hire moves the margin without moving freight — which is exactly the effect the frozen 2017 ratio cannot see.

## Source 2 — Freight Facts and Figures

https://data.bts.gov/stories/s/Freight-Transportation-the-Economy/6ix2-c8dn — BTS tables on freight transportation and the economy, including value of freight shipments and transportation services contribution to GDP. Not in bedrock; assess whether a table is worth an extractor or whether this is a one-off read for validation.

## What to compare

Per mode and year, 2017-2024, all indexed to 2017:

| series | source | already built? |
|---|---|---|
| control total per mode | #611 `control_total_components` | ✅ |
| for-hire output per mode | `BTS_TSA` | FBA exists, years short |
[...]

Issue #612, opened 2026-08-07
Step 4c-3: port the AWTS/ARTS annual trade margin control totals


WesIngwersen:
**Nowcast plan Step 4c, phase 3.** Parent: #571. Plan: [`margins_estimation_plan.md`](https://github.com/cornerstone-data/bedrock/blob/nowcast/bedrock/analysis/nowcasting/margins_estimation_plan.md).

Annual control totals for wholesale and retail margin. **This supplies the *level*, not the commodity allocation** — the allocation comes from #610's 2017-anchored rates.

## What exists

`Gross_Margins_2017.yaml` on the flowsa **`margins`** branch, already built, pulling:

- **`Census_AWTS`** — Annual Wholesale Trade Survey, NAICS `4231`–`4249`
- **`Census_ARTS`** — Annual Retail Trade Survey, NAICS `441`–`454`

Both give **gross margins, gross margin as a percent of sales, and sales**, annually for **2012–2022**. That is the trade-sector margin level, refreshed each year.

## Why the level is the easy half

The negative side of the margin columns is close to a function of trade-sector output, which Step 4a (#570) nowcasts anyway:

- **19 trade commodities** supply the whole −3,264,931, giving up **96.8%** of their combined `T007`. Eight are at **exactly −100.0%**: `441000` motor vehicle dealers, `445000` food and beverage stores, `452000` general merchandise, `448000` clothing, `444000` building material, `446000` health and personal care, `447000` gasoline stations, `4B0000` all other retail.
- **5 transport commodities** supply −415,570, at **56.8%** of theirs — lower because transport also sells directly.

A retail sector's output essentially *is* margin. So AWTS/ARTS plus 4a should pin the level well; the difficulty is entirely on the positive side.

## The work

1. Port `Gross_Margins_2017.yaml` and the `Census_AWTS` / `Census_ARTS` FBAs into bedrock's extract path.
2. Reconcile the AWTS/ARTS trade-sector totals against the 2017 Supply negatives, so the annual series is known to be on the same basis as the benchmark before it is used to move anything.
3. Wire as the control totals that #613 scales to.

## Note on scope

BEA's own method allocates trade margin to commodities via Economic Census product lines and the MAKB rule. Most of that machinery also exists (see the deferred issue on re-running BEA's product-line method), but it is **not** needed for this issue and probably not for v0.5 — #610's 2017 rates approximate its output. This issue is only the annual level.

WesIngwersen:
[...]

Issue #611, opened 2026-08-07
Step 4c-2: port the FAF transport margin chain from flowsa margins


WesIngwersen:
**Nowcast plan Step 4c, phase 2.** Parent: #571. Plan: [`margins_estimation_plan.md`](https://github.com/cornerstone-data/bedrock/blob/nowcast/bedrock/analysis/nowcasting/margins_estimation_plan.md).

**Transport is the one margin solved end-to-end by an external source** — FAF gives both the level and the commodity allocation. Most of this already exists and needs porting, not writing.

## What exists

| Piece | Where | State |
|---|---|---|
| `BTS_FAF.py`, `BTS_FAF.yaml` | flowsa **`margins`** branch | Built |
| `Transport_Margins_2017.yaml` | flowsa `margins` | Built — already selects `Unit: ton-miles` |
| `NAICS_Crosswalk_FAF_Mode_and_SCTG.csv` | flowsa `margins` | Built |
| `FAFData.R`, `Crosswalk_SCTGtoBEA.csv`, mode→BEA crosswalks | `cornerstone-data/stateior` | Built, **R** |

FAF 5.6.1 covers **2018–2023 annually** (FAF 4.5.1 back to 2012), which is the nowcast window — a real improvement on the manual's Commodity Flow Survey, which is quinquennial.

## Why FAF's mode coverage is exactly right

The five freight modes map onto **100% of the `TRANS` suppliers**, and nothing is missed:

| FAF mode | BEA | 2017 `TRANS` |
|---|---|---:|
| Truck | `484000` | −281,589 |
| Rail | `482000` | −68,590 |
| Pipeline | `486000` | −49,660 |
| Water | `483000` | −9,506 |
| Air | `481000` | −6,225 |
| | | **−415,570 = all of it** |

The other three modes — Multiple Modes and Mail (`491000`), couriers (`492000`), Other/Unknown (`48A000`) — map to commodities with **zero** `TRANS`, because postal and courier freight is bought as a direct service rather than embedded as margin. `Transport_Margins_2017.yaml` already handles this, redistributing them proportionally to the primary modes for the same commodity.

## The actual work

1. **Port `BTS_FAF` into bedrock's extract path** from flowsa `margins`.
[...]

Issue #610, opened 2026-08-07
Step 4c-1: derive the 2017 margin rates and the three receiving sets


WesIngwersen:
**Nowcast plan Step 4c, phase 1.** Parent: #571. Plan: [`margins_estimation_plan.md`](https://github.com/cornerstone-data/bedrock/blob/nowcast/bedrock/analysis/nowcasting/margins_estimation_plan.md).

**Depends on nothing — start here.** This is the half of 4c that is genuinely parallel, and doing it first de-risks the three downstream consumers (Supply margin columns, Step 6b PUR→PRO, Step 6d deliverable) without waiting on 4a or 4b.

## What to build

From the published 2017 detail Margins table (56,619 rows, 421 buyers × 404 commodities):

1. **The three receiving sets** — which `(buyer, commodity)` transactions bear each margin, from non-zero values:

   | margin | transactions | commodities |
   |---|---:|---:|
   | Wholesale | 17,027 (30.1%) | 253 |
   | Retail | 1,933 (3.4%) | 188 |
   | Transportation | 12,252 (21.6%) | 257 |

2. **Rates per `(buyer, commodity, margin type)`** on BEA's cascading bases — wholesale on `Producers' Value + Transportation`, retail on `Producers' + Transport + Wholesale`.

3. **Proof that the two identities reproduce the published Supply columns**, commodity by commodity:

   ```
   Σ_buyers ( Wholesale + Retail ) = TRADE[c] + TOP[c]
   Σ_buyers   Transportation       = TRANS[c]
   ```

   Current state: `TRANS` holds for 320/402 commodities within 1%; the trade identity **with** the `TOP` term holds for 226 of the 255 with positive `TRADE`.

## Three things that will bite

**Do not collapse to one rate per commodity.** Tested: zero of 249 wholesale commodities and zero of 56 retail commodities have a uniform rate (median CV 0.238 and 0.696). BEA's uniform rate is at the *item* level, finer than published commodities, and not recoverable.

**Exclude `F03000` when fitting rates.** All 31 negative-margin rows are change-in-inventories. A negative margin over a change-in-inventories base is a timing correction, not a rate, and will poison the fit. Carry its margin as a level instead. Do not clip or absolute the values — ⚠️ `_margin_negatives_treatment`'s `abs_negative_margin_columns` flag in `derive_PRO_to_PUR_ratio.py` would destroy the signal.

[...]

PR #627, opened 2026-08-10
Step 4c-2: port the FAF transport margin chain, annual TRANS 2017-2024 (#611)

WesIngwersen:
cc:
Related: #571, #611

**Stack 2 of 3** for Step 4c margins. Base #610 (`margins_2017_anchor`) — review that first.
→ #612 trade levels.

## What changed? Why?

Phase 2: the annual transportation margin column, 2017–2024. `nowcast_transport_margins.py` reproduces the published 2017 `TRANS` **to the dollar** and moves it forward on FAF ton-miles.

**Ported:** `BTS_FAF` extractor (FAF5.7.1), the mode/SCTG crosswalk, and `Margins_Transport_<year>` FBS methods for 2017–2024 sharing one `Margins_Transport_source.yaml`, so a year is eight lines.

Two departures from the flowsa original, both forced by the data:
- **No `tabula`/PDF scrape** — the SCTG and mode code tables ship inside the archive as `FAF5_metadata.xlsx`.
- **All trade types kept.** FAF's `dms_*` fields already describe only the domestic leg, so the `trade_type == 1` filter was discarding 20% of ton-miles, concentrated in imports, rather than avoiding a double count.

Two fixes the issue did not scope: `faf_parse` filtered on a year that `call_all_years: True` never passes it, so **no FBA could be generated at all**; and `BTS_FAF` was missing from `source_catalog.yaml`, without which any FBS reading it fails on `data_format`.

## The construction: ton-miles *move* the 2017 allocation, they do not generate it

Measured on 2017, allocating each mode's margin on its share of the measure — BEA's own construction — and scoring against published `TRANS` at SCTG level:

| source | measure | pearson | total abs share error |
|---|---|---:|---:|
| **FAF** | **ton-miles** | **0.370** | **68.2pp** |
| FAF | tons | 0.099 | 92.9pp |
| CFS PUF | ton-miles | 0.099 | 79.9pp |

**Ton-miles beat tons decisively**, so the manual's constant-revenue-per-ton assumption is the weaker one — weight-based allocation over-pays bulk (coal 35,609 predicted against 13,432 published) and starves light high-value goods (furniture 2,502 against 21,035).

**The CFS public use file is disqualified on scope**: it carries 2,979 of FAF's 5,436 billion ton-miles, and **crude petroleum, the largest `TRANS` commodity at 47,613, has zero CFS coverage.**

But neither measure *predicts* the margin, so the weak cross-sectional fit is made irrelevant rather than blocking — each commodity's published 2017 `TRANS` is moved by its SCTG group's ton-mile growth, controlled to the annual total. Exact in 2017 by construction, so BEA's own commodity allocation is inherited rather than approximated.

## The control total is most of the movement, and it is three treatments

⚠️ Ton-miles are a **volume** index and `TRANS` is **nominal**. The volume index reaches only 1.036 by 2022 while the level reaches 1.548 — nearly all movement is repricing, which the volume index cannot see. Truck is 68% of `TRANS` and its output runs 1.572× its 2017 level in 2022.

A single treatment cannot serve all five modes, because they differ in how much of their output is margin at all — 0.026 for air against 0.862 for rail:

| mode | share of `TRANS` | margin ÷ output | leverage | treatment |
|---|---:|---:|---:|---|
| truck | 67.8% | 0.767 | 0.30 | `residual` |
| rail | 16.5% | 0.862 | 0.16 | `residual` |
| pipeline | 11.9% | 1.033 | — | `output_ratio` |
| water | 2.3% | 0.173 | 4.8 | `freight_volume` |
| air | 1.5% | 0.026 | 36.9 | `freight_volume` |

A residual for air **goes negative** from 2019 (−121,719 million by 2024), and a frozen ratio would make air *freight* margin follow air *passenger* output, cutting it 46% in 2020 when air freight ton-miles were 1.010.

⚠️ **This choice is the largest open uncertainty in the column** — all four constructions are identities in 2017 and spread **11.8% by 2022** (`mixed` 643,443 against `freight_volume` 575,587). So the method is a **parameter**, not baked in: `control_total_components(years, method=…)` and `control_total_comparison(years)`. #620 settles it against external sources.

## Limitations recorded rather than corrected

- **The 2017/2018 seam is partly methodological** — FAF's 2017 is CFS-anchored, 2018 onward modelled. The `TRANS`-weighted mean step is 1.124 against the aggregate's 1.019, so a shape shift survives the control total.
- **The residual freezes the intermediate:final split at 2017.** Taking it from a nowcast Use matrix would be circular — Step 6b needs these margins to build it.
- ⚠️ **`BEA_Detail_GrossOutput_IO` is *industry* output; the margin given up is a *commodity* quantity.** For truck they differ by a third (367,154 against 333,879). The conclusion is unaffected, but the build should move to `T013` once 4a lands so the residual is one framework throughout.
- **The industry price index needs no commodity conversion — measured.** The transport block of the Make table is near-diagonal, so re-weighting moves the indices by at most 0.696% anywhere and changes the built control total by **0.000% and 0.084%**. This would not generalise to the trade commodities.

## Known limitation: the FBS cannot express BEA codes

⚠️ `SectorConsumedBy` resolves to NAICS, which is the #546 limitation — an SCTG mapped to a 3-digit NAICS group would be spread across everything in it. The module therefore aggregates the FBS back over its sector columns to the SCTG surviving in `Flowable` and joins to BEA detail on the crosswalk's `Note` column in Python. When #546 lands a BEA-code target the join moves into the FBS. For the same reason flowsa's `move_SCB_to_flow` is deliberately **not** ported: it overwrites `Flowable` with the NAICS sector and would silently coarsen every growth factor.

## Testing

`bedrock/transform/iot/__tests__/test_nowcast_transport_margins.py` (301 lines). The 2017 reproduction is exact.

Depends on `retain_activity_columns` from #601, confirmed present on `nowcast`.

### DIFF
--- a/bedrock/analysis/nowcasting/margins_estimation_plan.md
+++ b/bedrock/analysis/nowcasting/margins_estimation_plan.md
@@ -140,7 +140,107 @@ approximated, and the commodity cost structure a volume weighting discards is
 kept — furniture stays eight times dearer per ton-mile than coal because the
 2017 table says so, not because a model inferred it.
 
-Three things that follow, all of which shrink the work:
+### The control total is not decoration — it is most of the movement
+
+⚠️ **Ton-miles are a volume index and `TRANS` is nominal.** Moving 2017 by
+ton-mile growth alone holds 2017 freight *rates* fixed, and 2017 freight rates
+are not a small thing to get wrong: the volume index reaches only 1.036 by 2022
+while the nominal level reaches 1.483. Nearly all of the movement in the
+transport margin over the nowcast window is repricing, not more freight.
+
+**Built** in
+[`nowcast_transport_margins.py`](../../transform/iot/nowcast_transport_margins.py)
+as the transport industries' own gross output, held at each mode's 2017 ratio of
+margin given up to output:
+
+```
+control[t]   = Σ_m GO_m[t] × (given_up_m[2017] ÷ GO_m[2017]),  rescaled so control[2017] = TRANS_2017
+TRANS[c, t]  = shape[c, t] × control[t] ÷ Σ_c shape[c, t]
+```
+
+Two reasons for that source over a freight price index. It is **the same number
+the negative side of the column carries** — `Σ_m given_up_m` *is* `−TRANS`, so
+the two sides cannot drift apart — and it is per mode, so it reprices truck
+freight with truck output rather than with an economy-wide deflator. The
+rescaling is 0.24%: the 2017 composite is 415,548 against 414,559 received, the
+gap being the negative `F03000` rows on the receiving side.
+
+| | 2018 | 2019 | 2020 | 2021 | 2022 | 2023 | 2024 |
+|---|---:|---:|---:|---:|---:|---:|---:|
+| volume (ton-miles) | 1.066 | 1.052 | 0.997 | 1.025 | 1.036 | 1.029 | 1.032 |
+| level (control) | 1.099 | 1.127 | 1.089 | 1.266 | 1.483 | 1.393 | 1.399 |
+| **repricing** | 1.031 | 1.072 | 1.092 | 1.236 | **1.431** | 1.354 | 1.356 |
+
+Truck is 68% of `TRANS` and its output runs 1.572x its 2017 level in 2022 — the
+freight-rate surge, which the volume index cannot see at all.
+
+Three limitations, recorded rather than corrected:
+
+- **The 2017/2018 seam is partly methodological.** FAF's 2017 is anchored on the
+  CFS and 2018 onward are modelled. Aggregate ton-miles move 1.9% across it, but
+  the composition moves more: 21% of 2017 `TRANS` sits in SCTG groups jumping
+  more than 25% in that one year, and the `TRANS`-weighted mean step is 1.124
+  against the aggregate's 1.019. The control total absorbs the level, so what
+  survives is a shape shift.
+- **Air and rail output are not freight-only.** Air's 2017 give-up ratio is
+  0.026 against rail's 0.862 — the rest of air output is passengers — so air
+  freight margin follows passenger air. Air is 1.5% of `TRANS`; truck, the mode
+  that matters, is close to freight-pure.
+- **Pipeline gives up 1.033 of its own gross output.** Carried as an index, but
+  it means pipeline margin tracks the crude-price component of pipeline output.
+  `211000` crude petroleum is on that account the largest single movement in the
+  whole column, +34,289 million from 2017 to 2024.
+
+External validation of the annual trend — BTS Transportation Satellite Accounts,
+which give **for-hire** output by mode, and Freight Facts and Figures — is
+[#620](https://github.com/cornerstone-data/bedrock/issues/620), medium priority.
+`BTS_TSA` is already half-built as an FBA (2012–2019, NAICS 2012).
+
+### ⚠️ Agreed refinement, designed and not yet built
+
+The frozen give-up ratio is the weak point above, and there is a better
+construction that removes it for most of the column. **Decided 2026-08-09; the
+code still carries the ratio version.**
+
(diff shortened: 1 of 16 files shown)


WesIngwersen:
## Closing: BEA's transport margin method is not what this implements

BEA replied to the method question recorded in `7488437` (W. Nicolls, Section Chief for distributive services, 2026-08-11). The allocation basis in this PR is wrong for the two largest modes, so the chain is being withdrawn rather than patched.

### What BEA actually does

| mode | basis | source | frequency |
|---|---|---|---|
| **Rail** | **revenue by product** | AAR Freight Commodity Statistics (purchased); STB Commodity Revenue Stratification Report | annual |
| **Truck** | **revenue by product** | Service Annual Survey Table 8, ~sector level | annual |
| **Water** | ton-miles **× difficulty multiplier 1/2/3** | BTS/Census CFS | multipliers refreshed every 5 yrs |
| **Air** | ton-miles × multiplier (animals & fish only) | BTS/Census CFS | as above |

This PR allocates every mode on FAF ton-miles. That is approximately right for water and air (modulo the missing multiplier layer) and simply the wrong quantity for rail and truck.

The water multipliers encode how hard cargo is to rearrange at port: free-flowing bulk such as grain or oil stays at 1, palletized goods get 2, heavy machinery 3.

### What survives

Two architectural choices are confirmed, not overturned:

- **Shape/level separation.** "We really aren't looking for a specific value since that is already measured in margin output, we are simply trying to estimate the percentage of that margin output that is distributed to each commodity." That is exactly the construction in `70791c9`.
- **Strict per-mode treatment.** "We do not combine modes and we ignore multi-modal reported data since we cannot differentiate those in margin output." Confirms `7cf19ad`.

The `BTS_FAF` extractor fix (`b61cf98`) and `NAICS_Crosswalk_FAF_Mode_and_SCTG.csv` also remain useful, in the narrower role of allocating water and air. Both will be recovered from this branch rather than rewritten.

### It also explains the residual we could not

Q3 asked whether the published margin includes local delivery and small-shipment cost, which would explain this reconstruction under-allocating to light high-value commodities by 5–8×. BEA says no: no separate handling, and "the specific cost of a commodity doesn't figure in to the calculation, just the cost to ship it reported in the data."

The answer is elsewhere in the same reply. Rail and truck are allocated on **revenue**, and revenue per ton-mile is far higher for light high-value freight than for bulk. A pure ton-mile allocator starves exactly those commodities, by roughly the ratio of their freight rate to the bulk rate. The 5–8× gap was never a missing distribution-cost term — it was the wrong allocator applied to most of the value.

That in turn undermines how open question 1 was closed. Ton-miles beat tons on the 2017 aggregate fit, but that fit pooled four modes with different mechanisms, and for the two largest the right answer is not a volume measure at all. **Question 1 reopens**; §Transportation and the "Transport is solved end-to-end by FAF" claim both need rewriting.

### Sourcing for the rebuild

- **Rail** — AAR is proprietary, so that door is closed; the **STB Commodity Revenue Stratification Report is public** and is the substitute BEA names itself.
- **Truck** — SAS Table 8 is public but sector-level, which bounds achievable detail here *and* bounds BEA's own.
- **Water / air** — CFS ton-miles, public. Worth noting the FAF-over-CFS decision was taken for an all-mode allocator that no longer exists, and the "CFS has zero coverage of the largest `TRANS` commodity" objection may not bite once CFS is only carrying water and air.
- **The 1/2/3 multipliers are published nowhere.** Without that table water is approximable, not reproducible. Nicolls invited follow-up, so it is worth requesting directly, along with how the sector-level truck revenue is spread within sector.

#628 has been rebased onto #626 and is unaffected — it is trade margins, a different problem. #626 is untouched: it reads BEA's published 2017 margin table and involves no allocation method.

Branch retained at `faf_transport_margins` (`b94913f`); nothing is deleted.

Issue #624, opened 2026-08-10
API keys are only read from API_Keys.env, not the project .env


WesIngwersen:
`load_env_file_key` looks for API keys in exactly one place — `bedrock/extract/API_Keys.env`, naming each key by the `api_name` from the FBA config:

```python
if env_file == 'API_Key':
    load_dotenv(f'{extractpath}/API_Keys.env', verbose=True)
    value = os.getenv(key)          # key == 'Census', 'BEA', ...
    if value is None:
        raise APIError(api_source=key)
```

This project keeps its keys in the **root `.env`** under the conventional `<API_NAME>_API_KEY` spelling — `CENSUS_API_KEY`. Those two conventions do not meet, so every source with `api_key_required: true` raises

```
APIError: Key file Census not found. See README for help
```

on a machine that is otherwise correctly configured. The workaround is to copy the same secret into a second file under a second name, which is how secrets drift apart.

## Suggested fix

Accept both layouts, `API_Keys.env` first so nothing existing changes behaviour:

```python
load_dotenv(f'{extractpath}/API_Keys.env', verbose=True)
load_dotenv(MODULEPATH.parent / '.env', verbose=True)
value = os.getenv(key) or os.getenv(f'{key.upper()}_API_KEY')
```

`load_dotenv` does not overwrite variables already in the environment, so an exported key still wins and CI is unaffected. Both files are covered by the `*.env` line in `.gitignore`.

## Why it is worth fixing rather than documenting

The Census API does not answer a keyless request with an error status. It returns **HTTP 200 and an HTML "Missing Key" page**, so a request that silently lost its key fails inside the parser — as a JSON decode error, or worse, as a well-formed but wrong frame. The key being reliably found is what keeps that failure mode out of reach.

Found while porting `Census_AIES` for the trade margin control totals (#612), which is the first API-key source the nowcast path needs. The affected sources are every FBA config carrying `api_key_required: true` — `Census_CBP`, `Census_ACS`, `Census_ASM`, `Census_QWI`, `Census_PEP_Population`, `BEA_PCE`, `BLS_QCEW` and others.

Also worth deciding as part of this: the README the error message points at documents only the `API_Keys.env` layout, so it needs a line either way.

Issue #613, opened 2026-08-07
Step 4c-4: apply rates to nowcast years and derive TRADE/TRANS


WesIngwersen:
**Nowcast plan Step 4c, phase 4.** Parent: #571. Plan: [`margins_estimation_plan.md`](https://github.com/cornerstone-data/bedrock/blob/nowcast/bedrock/analysis/nowcasting/margins_estimation_plan.md).

Assemble the nowcast-year Margins table and derive the Supply columns from it.

## Depends on 4a *and* 4b — this is the sequencing constraint

BEA distributes margins on **total product supply including imports**:

```
T013 = T007 + MCIF + MADJ
```

So applying rates needs `T007` from **#570 (4a)** and `MCIF`/`MADJ` from **#579 (4b)**. Three independent statements confirm the base:

- *"Total product supply (column OR) includes total commodity output plus imports; margins are distributed based on the value in column OR."* — B. Jolliff, BEA, 2025-06-16
- The 2009 manual's wholesale step (4): product-line margin distributed to I-O commodities using **interim supply** as weights
- The manual's transport step (2), table 8.2: column C is **"INTERIM SUPPLY (A+B)"** = output + imports

⚠️ **Applying rates to `T007` alone to dodge the 4b dependency is the tempting shortcut and is wrong** — it drops the domestic margin on imported goods, which BEA explicitly includes ("the margin on imported goods will be included in both of the tables we discussed").

## Inputs

| from | what |
|---|---|
| #610 | 2017 rates per `(buyer, commodity, margin type)` and the receiving sets |
| #611 | FAF commodity × mode transport allocation, annual |
| #612 | AWTS/ARTS annual wholesale and retail levels |
| #570, #579 | `T007`, `MCIF`, `MADJ` → `T013` |

## Assemble

1. Apply #610's rates to nowcast `T013` on the receiving sets.
2. Control transport to #611, wholesale and retail to #612.
3. Carry `F03000` change-in-inventories margin as a **level**, not a rate — and expect the sign to flip in an inventory-build year. A nowcast that hard-codes the 2017 sign breaks when the cycle turns.
4. Derive the Supply columns, remembering the tax term:

   ```
   Σ_buyers ( Wholesale + Retail ) = TRADE[c] + TOP[c]
   Σ_buyers   Transportation       = TRANS[c]
   ```

   Dropping the `TOP` term overstates `TRADE` by 385,283 million, 11.8%.

[...]

Issue #571, opened 2026-08-05
Step 4c: transaction-level margins dataset, and the TRADE/TRANS columns from it


WesIngwersen:
**Nowcast plan Step 4c.** **Highest fan-out of any unbuilt piece in the project** — it feeds the Supply table's margin columns (4c), the PUR→PRO conversion (6b), *and* the Margins deliverable (6d). Building it early collapses three unknowns; building it late blocks all three.

Plan: [`.claude/plan/nowcast_plan.md`](https://github.com/cornerstone-data/bedrock/blob/nowcast/.claude/plan/nowcast_plan.md) §Step 4c.

**Runs in parallel with the BEA code-space work (#566-#569)** — transform-shaped, not FBS-shaped, so not gated by it.

## Build the fine object, derive the coarse one

The Margins table is per **commodity-buyer transaction**:

- index `(Industry Code, Commodity Code)`, where the industry side spans both `USA_2017_INDUSTRY_CODES` **and** `USA_2017_FINAL_DEMAND_CODES` — i.e. the buyer, intermediate or final
- columns `Producers' Value, Transportation, Wholesale, Retail, Purchasers' Value`
- BEA detail granularity

(see `bedrock/extract/iot/io_2017.py:333-359`)

The Supply table's `TRADE`/`TRANS` are single values per commodity. They reconcile by aggregation:

```
TRADE[c] = Σ_buyers ( Wholesale[b,c] + Retail[b,c] )
TRANS[c] = Σ_buyers   Transportation[b,c]
```

**Build the transaction-level dataset and derive the Supply columns from it.** The reverse is impossible — commodity totals cannot be disaggregated back to buyers without reintroducing the very assumption the Margins table exists to carry.

## Why transaction level is not optional

Step 6b's PUR→PRO conversion **requires** the buyer dimension: margin rates differ by buyer for the same commodity — a good bought by a household carries retail margin that the same good bought as an intermediate input may not — so a commodity-average rate would misallocate across columns.

USEEIO issue #4 flags that the external `nipa_final_demand_estimates.py` does this conversion with PCEBridge's 5 value-chain columns, which only covers PCE, not all of final demand, and leaves "splitting a margin category across individual commodity sectors" unresolved. **A transaction-level Margins dataset is exactly the fix**, since the buyer dimension is what PCEBridge was standing in for.

## Method

Start from 2017 detail margin **rates** per (buyer, commodity, margin type), carried forward on trade/transport output and commodity inflation.

[...]

Issue #564, opened 2026-08-05
Step 3: Probe annual business-survey expense data as a source for annual intermediate requirements


WesIngwersen:
**Optional / not on the critical path.** #497 delivers Step 3 without this; treat it as an improvement pass on top.

## Why

The current method for the SUT Use intermediate block (#497) seeds from the 2017 detail Use table and carries it forward on commodity inflation. That means every industry's **input structure is frozen at its 2017 shape** for the whole 2018-2025 span, and only the price level moves.

The Census annual business surveys — Manufactures, Retail, Wholesale, Services, Transportation — collect **inputs to production** annually. Critically, business expenses are reported at **what the firm actually paid**, i.e. purchaser prices, which is exactly the SUT Use cell basis. So they drop into the Use table's valuation with no conversion step, unlike most alternatives. They are also establishment-based and NAICS-coded, the axis bedrock already bridges to BEA Detail.

This issue is to **probe whether that data is good enough, and deep enough, to estimate annual requirements for the more specific industries in the Use table.**

## Desk research

Written up in [`bedrock/analysis/annual_survey_inputs/annual_survey_expense_sources.md`](https://github.com/cornerstone-data/bedrock/blob/nowcast/bedrock/analysis/annual_survey_inputs/annual_survey_expense_sources.md). Summary of what's already established:

- **The surveys were consolidated.** AIES replaced seven annual surveys (incl. ASM, SAS, ARTS, AWTS) starting with **data year 2023** — so our 2018-2025 span straddles the changeover. Census appears to have backfilled a harmonized **1992-2023** time series, which would remove the seam, but that needs verifying.
- **`timeseries/aies/exp02`** ("Detailed Operating Expenses") publishes **~35 named expense categories** in $1,000, by NAICS, **with a published coefficient of variation per cell**. Roughly 20 map plausibly to BEA Detail commodities (electricity, fuels, freight transportation, advertising, professional/technical, data processing, communications, software, repairs, rentals, insurance, temp staffing, …).
- **Some items belong to other steps**, not Step 3: payroll/fringe → `V00100`, taxes and license fees → `T00OTOP` (Step 2); cost of resale and purchased freight → margins (Step 4c). One pull informs three steps — and ownership needs deciding so two steps don't source the same dollars.
[...]

PR #628, opened 2026-08-10
Step 4c-3: annual wholesale and retail trade margin levels (#612)

WesIngwersen:
cc:
Related: #571, #612

**Stack 3 of 3** for Step 4c margins. Base #611 (`faf_transport_margins`) — review #610 and #611 first.

⚠️ Also needs #624 (`api_key_from_project_env`) for `Census_AIES` to read `CENSUS_API_KEY` from the project `.env`. Separate PR, no git dependency.

## What changed? Why?

Phase 3: the annual **level** of the trade margin, 2012–2023. No commodity dimension — #610's 2017-anchored rates do the allocation; this only sizes the total being allocated.

**Three extractors, not the two the issue scoped.** From data year 2023 the Census Bureau consolidated its annual economic surveys into the Annual Integrated Economic Survey, so AWTS and ARTS stop at 2022 and `Census_AIES` carries 2023 from `timeseries/aies/basic`. The issue was written against flowsa's 2012–2022 picture and did not know this.

⚠️ **The splice only works at the right type of operation.** AIES publishes wholesale gross margin under `TYPOP 1X` (merchant wholesalers excluding MSBOs) and **zero** under the all-types code `00`; retail is the other way round. The wrong code returns a well-formed zero rather than an error, silently deleting one side of the trade margin — so the codes are pinned per sector and an empty sector raises.

## These are index numbers, not levels

| 2017, $M | Census survey | BEA published column | ratio |
|---|---:|---:|---:|
| Wholesale, NAICS 42 row | 1,100,925 | 1,894,329 | **0.581** |
| Retail, `Total` row | 1,458,243 | 1,761,765 | **0.828** |

So Phase 3 hands Phase 4 a **growth factor on the 2017 BEA level**, the same anchor-and-move construction Phase 2 used for transport. Substituting the Census level directly would delete 42% of the wholesale margin.

## Decision: anchor on the annual survey, not the Economic Census

The EC is more authoritative in general, and that is the right reason to expect it to win a level dispute — but it is not the criterion here. Phase 3 supplies **movement only**; the level comes from the 2017 BEA column by construction. The EC is quinquennial, so it cannot supply annual movement at all, and interpolating 2012/2017/2022 would add modelled years dressed as observations. The 42% EC-vs-AWTS gap therefore never enters the build — it cancels in each year's ratio **provided it is stable over time**, which is the assumption now doing the work, so it was tested where it bites.

## The splice is the seam that matters, and only one side is clean

On the published total rows, the margin *rate* is the test, being invariant to coverage:

| gross margin ÷ sales | 2020 | 2021 | 2022 | **2023 (AIES)** | step |
|---|---:|---:|---:|---:|---:|
| wholesale, NAICS 42 | 19.7% | 20.7% | 20.1% | **20.4%** | **+0.3pp** ✅ |
| retail, `Total` | 29.5% | 30.8% | 31.3% | **34.2%** | **+2.9pp** ⚠️ |

**Wholesale splices cleanly** — level −1.0% on sales −2.5%, an ordinary 2023 as goods inflation cooled. The `TYPOP 1X` = `nomsbo` basis match holds in the data, not just the documentation.

**Retail does not, yet.** The rate steps 2.9pp against a series that moved ~0.3pp/yr for a decade, on sales up only 1.5%. Not a NAICS vintage artifact: AIES 2023 publishes the 2017-vintage retail structure (`452` and `454` present, no `455`–`459`) and its twelve 3-digit codes sum to the published `44-45` total exactly. **The discriminating test** is recorded in the plan — 2022 is an EC year, so ARTS 2022 / `ecnmargin` 2022 / AIES 2023 together say whether to rebase 2023 or keep it. Prerequisite of #613's control total, not of these extractors.

## ⚠️ Read the published total row, never the sum of sub-industries

Suppression varies by year, so the two are not a fixed distance apart:

| 4-digit sum ÷ NAICS 42 row | 2012–16 | 2017 | 2018 | 2019–21 | **2022** |
|---|---:|---:|---:|---:|---:|
| wholesale | 1.000 | 0.948 | 0.941 | 1.000 | **0.839** |

That is a **17.8pp error in one year of the growth factor**: 2021→2022 reads −6.4% on the 4-digit sum where the published row moves +11.4%. Retail has the same hole in 2022 (0.940). A specification for #613, not a defect in the extract — both aggregates are emitted correctly.

## Two departures from flowsa's `Gross_Margins_2017.yaml`

- **`Purchases` is not relabelled as `Sales`.** Trade gross output excludes the cost of goods purchased for resale, and purchases *are* that cost; sales are roughly purchases plus the margin. Measured on 2017, gross margin over gross-margin-percent agrees with purchases plus margin to within 1% for every kind of business. The relabel understates sales by the margin itself and overstates any rate built on it by about a fifth.
- **The trade sector lands in `ActivityProducedBy`.** It *produces* the margin — the placement `Margins_Transport` gives the freight mode, and what lets #613 read `TRADE` and `TRANS` the same way.

## The 850,540 wholesale gap, and why it stays open

Chased to the bottom, because "add the missing types of operation" is the obvious move and it does not work. **Agents and brokers close 3.6% and no more** — they never take title, so they book a commission rather than a margin, and before 2023 their only source is the quinquennial EC. **MSBO margin is published nowhere in any vintage**: `ecnmargin`/`ecnprofit` carry only `TYPOP 10`, and `ecntypop` breaks MSBOs out for sales and payroll but carries no margin item. So the index stays merchant-wholesaler-only, which is the consistent basis a growth factor needs. One lookout is recorded in #564's companion doc, since that probe will be reading the manufacturing tables closely and an MSBO markup may be visible from the parent's side.

## Two gaps that are not bugs

- **2024 has no source.** AIES returns 204 for every year but 2023, so the last year of the nowcast window must be extrapolated until the next release.
- **Suppression subtracts from a control total.** 2022 retail sums to 2,036,590 against a published 2,167,261 — 130,671 short, from suppressed gasoline stations and `44812`. The `Suppressed` flag is preserved so those cells can be recovered by subtraction from the published total.


WesIngwersen:
Rebased onto `margins_2017_anchor` (#626) and retargeted, since #627 has been closed — BEA's reply on the transport allocation basis invalidates the FAF chain it carried ([why](https://github.com/cornerstone-data/bedrock/pull/627#issuecomment-5285229382)).

Nothing in this PR changes: same three commits, same nine files, 857 insertions and no deletions. Trade margins are a separate problem from transport and are untouched by the BEA finding. The only effect is that this no longer sits on top of a withdrawn chain.

Pre-rebase head was `6f21ee6` if anything needs recovering.

